PR #10, opened 2024-10-01
Prototype: Add initial code for IRv2 decoder.

junhaoliao:
The changes in this PR are only to facilitate discussions. The original IR decoder was modified directly for the ease of prototyping and to show the differences between the two decoders. (Maybe we would end up with only one decoder depending on our discussion results.)

# Description

# Validation performed


CMakeLists.txt
```diff
@@ -88,7 +88,10 @@ if(CMAKE_BUILD_TYPE MATCHES "Release")
         -sENVIRONMENT=worker
     )
 else()
-    set(CLP_FFI_JS_EXTRA_LINKER_FLAGS -sENVIRONMENT=node)
+    set(CLP_FFI_JS_EXTRA_LINKER_FLAGS
+        -sENVIRONMENT=node
+        -sNO_DISABLE_EXCEPTION_CATCHING
+    )
```
junhaoliao, line 94: The added flag enables easier debugging and is not directly related to this PR.

src/clp_ffi_js/ir/StreamReader.cpp
```diff
@@ -46,31 +46,8 @@ auto StreamReader::create(DataArrayTsType const& data_array) -> StreamReader {
     auto zstd_decompressor{std::make_unique<clp::streaming_compression::zstd::Decompressor>()};
     zstd_decompressor->open(data_buffer.data(), length);
 
-    bool is_four_bytes_encoding{true};
```
junhaoliao, line 49: I had to remove this byte-encoding check as it seems to consume bytes from the reader and cause below `clp::ffi::ir_stream::Deserializer::create(*zstd_decompressor)` to fail with a protocol error. Do we still need this byte-encoding check?

I haven't taken a careful look into the clp-core code. Just out of curiosity, are we sticking with 4-byte or 8-byte encoding in IRfv2 or do we support both? I recall there were some discussions to deprecate one in favour of the other but I don't remember we had a conclusion.

src/clp_ffi_js/ir/StreamReader.cpp
```diff
@@ -46,31 +46,8 @@ auto StreamReader::create(DataArrayTsType const& data_array) -> StreamReader {
...
-    }
-    if (false == is_four_bytes_encoding) {
-        throw ClpFfiJsException{
-                clp::ErrorCode::ErrorCode_Unsupported,
-                __FILENAME__,
-                __LINE__,
-                "IR stream uses unsupported encoding."
-        };
-    }
-
```
junhaoliao, line 71: I had to remove this byte-encoding check as it seems to consume bytes from the reader and cause below `clp::ffi::ir_stream::Deserializer::create(*zstd_decompressor)` to fail with a protocol error. Do we still need this byte-encoding check?

I haven't taken a careful look into the clp-core code. Just out of curiosity, are we sticking with 4-byte or 8-byte encoding in IRfv2 or do we support both? I recall there were some discussions to deprecate one in favour of the other but I don't remember we had a conclusion.

src/clp_ffi_js/ir/StreamReader.cpp
```diff
@@ -112,8 +89,9 @@ auto StreamReader::deserialize_range(size_t begin_idx, size_t end_idx) -> size_t
     if (nullptr != m_stream_reader_data_context) {
         constexpr size_t cDefaultNumReservedLogEvents{500'000};
         m_encoded_log_events.reserve(cDefaultNumReservedLogEvents);
+        auto &reader{m_stream_reader_data_context->get_reader()};
         while (true) {
-            auto result{m_stream_reader_data_context->get_deserializer().deserialize_log_event()};
+            auto result{m_stream_reader_data_context->get_deserializer().deserialize_to_next_log_event(reader)};
```
junhaoliao, line 94: Note: pull out the deserializer to reduce line length.

src/clp_ffi_js/ir/StreamReader.cpp
```diff
@@ -150,43 +128,19 @@ auto StreamReader::decode_range(size_t begin_idx, size_t end_idx) const -> Decod
...
-    constexpr size_t cDefaultReservedMessageLength{512};
-    message.reserve(cDefaultReservedMessageLength);
     size_t log_num{begin_idx + 1};
     auto const results{emscripten::val::array()};
     for (auto const& log_event : log_events_span) {
-        message.clear();
-
-        auto const parsed{log_event.get_message().decode_and_unparse()};
-        if (false == parsed.has_value()) {
+        auto const json{log_event.serialize_to_json()};
```
junhaoliao, line 134: Here we retrieve the log event as a JSON since I am concerned about the cost of inserting nodes into an Embind object when we traverse the log event. Let me know if there're any concerns that I missed.

junhaoliao: Some benchmarking might be worthy before we decide on that though.

junhaoliao:
@LinZhihao-723 / @kirkrodrigues 
In addition to the inline questions, could you help me with below queries?
1. Shall we plan for two decoder classes or a single one for IRv1 and IRv2?
   * If we're to have two classes
     - Shall we release the classes as separate binaries?
       - Note there could be size overhead in the binaries.
       - That likely means we need to differentiate IRv2 from IRv1 in the Log Viewer. Not sure about how yet.
   * If we wish to maintain a single class
     - Do we plan to have the KeyValuePair as a new encoding type? That way it can be passed as a template variable to classes.
     - How do we differentiate IRv2 from IRv1 though? Do we check the version number from the file? (I got a bit confused by the clp-core code as it seems the "beta" version is referring to IRv2 while IRv1 is the regular version. Could you confirm?)
2. I think there have been different iterations of this discussion though I'm not sure about the latest conclusion: What're the official names for IRv1 and IRv2?
   - How should we name the class / classes?

LinZhihao-723:
> @LinZhihao-723 / @kirkrodrigues In addition to the inline questions, could you help me with below queries?
> 
> 1. Shall we plan for two decoder classes or a single one for IRv1 and IRv2?
>    
>    * If we're to have two classes
>      
>      * Shall we release the classes as separate binaries?
>        
>        * Note there could be size overhead in the binaries.
>        * That likely means we need to differentiate IRv2 from IRv1 in the Log Viewer. Not sure about how yet.
>    * If we wish to maintain a single class
>      
>      * Do we plan to have the KeyValuePair as a new encoding type? That way it can be passed as a template variable to classes.
>      * How do we differentiate IRv2 from IRv1 though? Do we check the version number from the file? (I got a bit confused by the clp-core code as it seems the "beta" version is referring to IRv2 while IRv1 is the regular version. Could you confirm?)
> 2. I think there have been different iterations of this discussion though I'm not sure about the latest conclusion: What're the official names for IRv1 and IRv2?
>    
>    * How should we name the class / classes?

1. In my opinion we should maintain two classes for now before we have full supports of IR v2, including: (1) backward compatibility with IR v1; (2) metadata kv-pair supports. That's also why the version number for v2 is still marked as "beta". But due to my limited understanding of the new log viewer, I actually don't know the implications/overheads of the two alternatives you proposed. We can schedule an offline meeting if you want.
2. There's no special name for IR v1. For IR v2, I think a more formal name we used in our core PRs is "key-value pair IR format"

src/clp_ffi_js/ir/StreamReader.cpp
LinZhihao-723: For v2 deserializer, you don't need this check. In terms of encoding types, both are supported by the format. However, the logging library will probably only implement four-byte encoding interface. The deserializer will automatically handle both encodings

kirkrodrigues:
> 1. Shall we plan for two decoder classes or a single one for IRv1 and IRv2?

We should plan for one stream reader (a.k.a. decoder / deserializer) class by the time we start using this in production. Before now and then, we can do whatever is simplest.

> If we're to have two classes, shall we release the classes as separate binaries?

Do we need to? The way I see it, the two readers are part of the same IR library, so it feels easier to keep them in one binary.

> Note there could be size overhead in the binaries.

I guess it's still smaller than the IR streams we end up decoding?

> That likely means we need to differentiate IRv2 from IRv1 in the Log Viewer. Not sure about how yet.

Right, this is another reason to go with one binary.

> If we wish to maintain a single class, do we plan to have the KeyValuePair as a new encoding type? That way it can be passed as a template variable to classes.

The production release plan is that the KV-pair IR format will support deserializing IR v1 streams and expose them with the same API as KV-pair IR streams. So hopefully we can move away from all the different template specializations.

> How do we differentiate IRv2 from IRv1 though? Do we check the version number from the file? (I got a bit confused by the clp-core code as it seems the "beta" version is referring to IRv2 while IRv1 is the regular version. Could you confirm?)

In my mind, the production release will use a different version number to differentiate the two.

> I think there have been different iterations of this discussion though I'm not sure about the latest conclusion: What're the official names for IRv1 and IRv2?

As Zhihao said above.

> How should we name the class / classes?

`IRStreamReader` and `KVPairIRStreamReader`? I think reader is more appropriate than decoder / deserializer since these classes really do both. Hopefully we don't have to keep these names for long.

junhaoliao:
@kirkrodrigues Thanks for completing the responses. I'll stick with one binary and redesign the interfaces with one stream reader pictured.

I also shortly discussed offline with Zhihao yesterday, and we have these items opened for discussions (@LinZhihao-723 please correct me if I'm wrong):
1. Currently we need to select a correct deserializer depending on the IR file version, but there is no existing code in `clp-core` that can directly differentiate an IRv1 stream from an IRv2 one. We explored below options:
   1. Assign a different magic number to IRv2 files.
   2. `decode_preamble(reader)` in C++ code and read the version out, although that would consume bytes from the reader, and directly reusing the reader for deserializer creation would cause protocol failures. We can make a copy before passing it to `decode_preamble` though.
   3. Since the metadata is stored as a JSON in the preamble, "peeking" into the first few bytes in JS code to decode the version out before calling into C++ code is a possible option.
2. > KV-pair IR format will support deserializing IR v1 streams and expose them with the same API as KV-pair IR streams

   Do we have a timeline for this to be ready?

CMakeLists.txt
junhaoliao: This shall be done in #11 . Reverting the changes in this PR

junhaoliao, line 94:
```suggestion
    set(CLP_FFI_JS_EXTRA_LINKER_FLAGS -sENVIRONMENT=node)
```

junhaoliao:
```suggestion
    set(CLP_FFI_JS_EXTRA_LINKER_FLAGS -sENVIRONMENT=node)
```

src/clp_ffi_js/ir/StreamReader.cpp
```diff
@@ -28,199 +30,72 @@
...
-        EM_ASM(
-                { Emval.toValue($0).push([UTF8ToString($1), $2, $3, $4]); },
-                results.as_handle(),
-                message.c_str(),
-                log_event.get_timestamp(),
-                log_level,
-                log_num
-        );
-        ++log_num;
+        return std::make_unique<KVPairIRStreamReader>(KVPairIRStreamReader::create(std::move(data_array)));
```
junhaoliao, line 98: This seems a bad practice since we have to recreate the data_buffer in KVPairIRStreamReader's factory function. Shall we make KVPairIRStreamReader's constructor public / friend so that we can
1. seek(0) in the zstd reader
2. create the deserializer from the zstd reader and then the StreamReaderDataContext
3. call KVPairIRStreamReader's constructor here.

so long we don't expose the constructor via Embind there should be minor concerns of misuses of the constructor (at the moment we are not exposing StreamReaderDataContext anyways). What do you think?

junhaoliao:
The PR is not ready for review yet as I'm still cleaning up the log level filtering code for IRv2.

junhaoliao:
This prototype is for reference only and not meant to be merged.

junhaoliao:
Closing this since all features prototyped in this PR has been checked in.

PR #11, opened 2024-10-01
Enable exception catching; Separate build targets for `node` and `worker` environments.

junhaoliao:
# Description

1. Enable exception catching with flag [`-fwasm-exceptions`](https://emscripten.org/docs/porting/exceptions.html?highlight=no_disable_exception_catching#webassembly-exception-handling-based-support) for all build types (i.e., "Release", "Debug") for easier debugging.
2. Instead of compiling for environment "node" in the "Debug" build type and environment "worker" in the "Release" type, separate those into different targets, which allows optimization flags to be enabled with "Release" build type even when building for environment "node".
3. Update target names in Taskfile.yml .
4. Update package.json exports to export the release binaries for both environments.

# Validation performed

1. Built the js assets with `task` and ran below sample code:
    ```
    import ModuleInit from "./cmake-build-release/ClpFfiJs-node.js"
    import fs from "node:fs"
    
    const main = async () => {
        const file = fs.openSync("./empty-file", "w")
    
        console.time("perf")
        const Module = await ModuleInit()
        try {
            const decoder = new Module.ClpIrStreamReader(new Uint8Array(file))
            const numEvents = decoder.deserializeRange(0, 0)
            const results = decoder.decodeRange(0, numEvents)
            console.log(results)
        } catch (e) {
            console.trace("Exception caught:", e)
        }
        console.timeEnd("perf")
    }
    
    void main()
    ```
2. Observed below in the console:
    ```
    /path/to/.nvm/versions/node/v22.1.0/bin/node /path/to/clp-ffi-js/test.mjs
    [2024-10-06 03:35:43.545] [info] [StreamReader.cpp:37] StreamReader::create: got buffer of length=14
    [2024-10-06 03:35:43.553] [error] [Decompressor.cpp:100] streaming_compression::zstd::Decompressor: ZSTD_decompressStream() error: Unknown frame descriptor
    [2024-10-06 03:35:43.553] [critical] [StreamReader.cpp:55] Failed to decode encoding type, err=4
    Trace: Exception caught: Exception [WebAssembly.Exception] {}
        at main (file:///path/to/clp-ffi-js/test.mjs:15:17)
    perf: 19.961ms
    
    Process finished with exit code 0
    ```

## Summary by CodeRabbit

- **New Features**
	- Introduced support for multiple executable targets for different environments (node and worker).
	- Added new module structure with distinct entry points for each environment.

- **Chores**
	- Updated build configuration for improved flexibility and modularity.
	- Adjusted package configuration to reflect new module type and export paths.


CMakeLists.txt
```diff
@@ -88,7 +88,10 @@ if(CMAKE_BUILD_TYPE MATCHES "Release")
         -sENVIRONMENT=worker
     )
 else()
-    set(CLP_FFI_JS_EXTRA_LINKER_FLAGS -sENVIRONMENT=node)
+    set(CLP_FFI_JS_EXTRA_LINKER_FLAGS
+        -sENVIRONMENT=node
+        -sNO_DISABLE_EXCEPTION_CATCHING
```
kirkrodrigues, line 93: I haven't looked into it myself, but would [`-fwasm-exceptions`](https://emscripten.org/docs/porting/exceptions.html?highlight=no_disable_exception_catching#emscripten-javascript-based-exception-support) also work?

package.json
```diff
@@ -15,13 +15,22 @@
   "scripts": {
     "release": "git diff --exit-code && npm publish"
   },
+  "type": "module",
   "files": [
-    "./dist"
+    "/dist"
```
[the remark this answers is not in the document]
junhaoliao: I believe the `/` refers to the project root. Correct me if I'm wrong.

CMakeLists.txt
junhaoliao: Both [`-fexceptions`](https://emscripten.org/docs/porting/exceptions.html?highlight=no_disable_exception_catching#emscripten-javascript-based-exception-support) and [`-fwasm-exceptions`](https://emscripten.org/docs/porting/exceptions.html?highlight=no_disable_exception_catching#webassembly-exception-handling-based-support) work. 

Enabling exception catching with `-fwasm-exceptions` seems to introduce minor performance impacts. i.e. Decoding all log events in a sample log file with ~380k log events 5 times takes 10.486s without the flag and 10.504s with the flag.

On the other hand, with JS-emulated exceptions enabled by `-fexceptions`, decoding time increases significantly to 17.087s.

As a result, we should enable `-fwasm-exceptions`.

package.json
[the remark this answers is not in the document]
junhaoliao: The slash was needed to mitigate an issue when I was simulated the package publish / installation locally with Yalc on a Mac computer (previously I don't recall observing the issue on a Linux computer). With the "./", the "dist" folder does not seem to be correctly included in the installations. 
Still, it seems with no "./" or "/" added the installation and the config works fine locally with Yalc. Will stick with simply "dist" then.

Taskfile.yml
```diff
@@ -41,7 +41,7 @@ tasks:
       - "src/**/*"
     generates: ["{{.CHECKSUM_FILE}}"]
     deps:
-      - "emsdk"
+#      - "emsdk"
```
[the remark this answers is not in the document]
junhaoliao: Right, that was a typo. Reverted.

package.json
```diff
@@ -15,13 +15,22 @@
...
+      "import": "./dist/ClpFfiJs-worker.js",
+      "types": "./dist/ClpFfiJs-worker.d.ts"
+    },
+    "./node": {
+      "import": "./dist/ClpFfiJs-node.js",
+      "types": "./dist/ClpFfiJs-node.d.ts"
+    },
+    "./worker": {
+      "import": "./dist/ClpFfiJs-worker.js",
+      "types": "./dist/ClpFfiJs-worker.d.ts"
```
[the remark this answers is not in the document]
junhaoliao: The files are generated rather than directly checked in, so it's expected you don't see the files under /dist without any compilations.

CMakeLists.txt
```diff
@@ -148,11 +108,64 @@ set(CLP_FFI_JS_SRC_ZSTD
...
+    target_sources(
+        ${CLP_FFI_JS_BIN_NAME}
+        PRIVATE
         ${CLP_FFI_JS_SRC_MAIN}
         ${CLP_FFI_JS_SRC_CLP_CORE}
         ${CLP_FFI_JS_SRC_FMT}
         ${CLP_FFI_JS_SRC_ZSTD}
-)
+    )
+endforeach()
```
junhaoliao, line 171:
```suggestion
endforeach()

```

CMakeLists.txt
```diff
@@ -67,59 +67,19 @@ message(STATUS "Fetching Boost.")
...
-    CACHE STRING
-    "Binary name for the generated .js and .wasm files."
-)
-add_executable(${CLP_FFI_JS_BIN_NAME})
-
-target_compile_features(${CLP_FFI_JS_BIN_NAME} PRIVATE cxx_std_20)
-
 set(CMAKE_EXECUTABLE_SUFFIX ".js" CACHE STRING "Binary type to be generated by Emscripten.")
 
+set(CLP_FFI_JS_ENABLE_EXCEPTION_FLAG
```
kirkrodrigues, line 72: We have this, `CLP_FFI_JS_EXTRA_LINKER_FLAGS`, and a bunch more linker options in the foreach below. How about defining a single list variable `CLP_FFI_JS_EXTRA_LINK_OPTIONS` (to match `target_link_options`) and just append to it as necessary?

kirkrodrigues: If we end up keeping this option, how about `CLP_FFI_JS_ENABLE_EXCEPTIONS_FLAG`? (Plural exceptions.)

CMakeLists.txt
```diff
@@ -148,11 +108,64 @@ set(CLP_FFI_JS_SRC_ZSTD
     src/submodules/zstd/lib/decompress/zstd_decompress.c
 )
 
-target_sources(
-    ${CLP_FFI_JS_BIN_NAME}
-    PRIVATE
+# List of supported environments.
+set(CLP_FFI_JS_ENVIRONMENTS
```
kirkrodrigues, line 112: * How about `CLP_FFI_JS_SUPPORTED_ENVIRONMENTS`? Then we can drop the comment above.
* Can we make this a cache variable so that we can set it from the Taskfile? That way, we don't have to keep the environments here in sync with the Taskfile.

CMakeLists.txt
```diff
@@ -148,11 +108,64 @@ set(CLP_FFI_JS_SRC_ZSTD
...
+    "node"
+    "worker"
+)
+
+foreach(env ${CLP_FFI_JS_ENVIRONMENTS})
+    set(CLP_FFI_JS_BIN_NAME "ClpFfiJs-${env}")
+    add_executable(${CLP_FFI_JS_BIN_NAME})
+    target_compile_features(${CLP_FFI_JS_BIN_NAME} PRIVATE cxx_std_20)
+    target_compile_options(${CLP_FFI_JS_BIN_NAME} PRIVATE ${CLP_FFI_JS_ENABLE_EXCEPTION_FLAG})
```
kirkrodrigues, line 121: Do we need to [add](https://emscripten.org/docs/optimizing/Optimizing-Code.html?highlight=lto#lto) `-flto`?

CMakeLists.txt
```diff
@@ -148,11 +108,64 @@ set(CLP_FFI_JS_SRC_ZSTD
...
+
+    set(CLP_FFI_JS_LINKER_FLAGS
+        ${CLP_FFI_JS_ENABLE_EXCEPTION_FLAG}
+        ${CLP_FFI_JS_EXTRA_LINKER_FLAGS}
+        -sALLOW_MEMORY_GROWTH
+        -sEXPORT_ES6
+        -sMODULARIZE
+        -sWASM_BIGINT
+        --emit-tsd=${CLP_FFI_JS_BIN_NAME}.d.ts
```
kirkrodrigues, line 130: Should we also be specifying `sENVIRONMENT=${env}`?

CMakeLists.txt
```diff
@@ -148,11 +108,64 @@ set(CLP_FFI_JS_SRC_ZSTD
...
+    )
+    message(
+        "CLP_FFI_JS_BIN_NAME=\"${CLP_FFI_JS_BIN_NAME}\". \
+CMAKE_BUILD_TYPE=\"${CMAKE_BUILD_TYPE}\". \
+Linker flags: ${CLP_FFI_JS_LINKER_FLAGS}"
```
kirkrodrigues, line 135:
```suggestion
Extra link options: ${CLP_FFI_JS_LINKER_FLAGS}"
```

Taskfile.yml
```diff
@@ -56,7 +56,7 @@ tasks:
         Emscripten.cmake" \
         -S "{{.ROOT_DIR}}" \
         -B "{{.OUTPUT_DIR}}"
-      - "cmake --build '{{.OUTPUT_DIR}}' --parallel --target '{{.G_CLP_FFI_JS_TARGET_NAME}}'"
+      - "cmake --build '{{.OUTPUT_DIR}}' --parallel --target {{.G_CLP_FFI_JS_TARGET_NAME}}"
```
kirkrodrigues, line 59:
```suggestion
      - "cmake --build '{{.OUTPUT_DIR}}' --parallel --target {{.G_CLP_FFI_JS_TARGET_NAMES}}"
```

kirkrodrigues (changes requested):

CMakeLists.txt
kirkrodrigues: Also, does this need to be a CACHE variable?

CMakeLists.txt
junhaoliao: > Thus, to allow maximal LTO opportunities with the LLVM Wasm backend, build all source files with -flto and also link with flto.

It seems so

CMakeLists.txt
junhaoliao: My initial thoughts of creating this variable is to make this configurable. Now I think about it again it seems there's probably no reason to change the option to the JS-emulated one (`-fexceptions`) once we decide use one type of exception but not the other. On the other hand, now we have confirmed the effects of enabling `-fwasm-exceptions` are minor, we should probably always enable it given how the CLP core code depends on exceptions for error handling.

I agree that we should create a single list `CLP_FFI_JS_EXTRA_LINK_OPTIONS` and it makes sense for us to inline this exception flag into the list directly.

CMakeLists.txt
junhaoliao: yeah sorry for the oversight: https://emscripten.org/docs/tools_reference/emcc.html#emcc-closure:~:text=%5Bcompile%2Blink%5D%20Enables%20link%2Dtime%20optimizations%20(LTO).

CMakeLists.txt
junhaoliao: Sorry, we should. I think I mistakenly deleted it when I was merging from main.

CMakeLists.txt
```diff
@@ -58,59 +58,28 @@ message(STATUS "Fetching Boost.")
...
-    CACHE STRING
-    "Binary name for the generated .js and .wasm files."
-)
-add_executable(${CLP_FFI_JS_BIN_NAME})
-
-target_compile_features(${CLP_FFI_JS_BIN_NAME} PRIVATE cxx_std_20)
-
 set(CMAKE_EXECUTABLE_SUFFIX ".js" CACHE STRING "Binary type to be generated by Emscripten.")
 
+# Set up extra compile and link options and append to the lists as necessary.
```
junhaoliao, line 63:
```suggestion
# Set up common compile and link options to be merged with other options as necessary.
```

CMakeLists.txt
```diff
@@ -139,11 +108,63 @@ set(CLP_FFI_JS_SRC_ZSTD
...
+
+    # Set up compile options
+    target_compile_features(${CLP_FFI_JS_BIN_NAME} PRIVATE cxx_std_20)
+    target_compile_definitions(${CLP_FFI_JS_BIN_NAME} PUBLIC SPDLOG_FMT_EXTERNAL=1)
+    target_compile_options(${CLP_FFI_JS_BIN_NAME} PRIVATE ${CLP_FFI_JS_COMMON_COMPILE_OPTIONS})
+
+    # Set up link options
+    target_link_libraries(${CLP_FFI_JS_BIN_NAME} PRIVATE embind)
+    set(CLP_FFI_JS_LINK_OPTIONS
+        ${CLP_FFI_JS_COMMON_LINK_OPTIONS}
```
junhaoliao, line 129: I understand the previous suggestion was to make an "extra link options" list and append to it as necessary. However, I'm not sure if the appended options could be undesirably carried onto the next `foreach` iterations. Instead of creating an "extra options" list, a "common options" list is created to merge with other options as necessary. Let me know if you see any drawbacks with this approach.

Taskfile.yml
```diff
@@ -8,7 +8,11 @@ vars:
   G_BUILD_DIR: "{{.ROOT_DIR}}/build"
   G_CLP_FFI_JS_BUILD_DIR: "{{.G_BUILD_DIR}}/clp-ffi-js"
   G_CLP_FFI_JS_CHECKSUM: "{{.G_BUILD_DIR}}/clp-ffi-js.md5"
-  G_CLP_FFI_JS_TARGET_NAME: "ClpFfiJs"
+  G_CLP_FFI_JS_ENV_NAMES: "node;worker"
```
kirkrodrigues, line 11: This is not a big deal, but for clarity, I would rather make this a map and then assemble the `;`-delimited string in `config-cmake-project`.

CMakeLists.txt
```diff
@@ -139,11 +108,63 @@ set(CLP_FFI_JS_SRC_ZSTD
     src/submodules/zstd/lib/decompress/zstd_decompress.c
 )
 
-target_sources(
-    ${CLP_FFI_JS_BIN_NAME}
-    PRIVATE
+set(CLP_FFI_JS_SUPPORTED_ENVIRONMENTS
+    "node;worker"
```
kirkrodrigues, line 112:
```suggestion
    node
    worker
```

I prefer separating them on different lines but feel free to disagree.

CMakeLists.txt
```diff
@@ -139,11 +108,63 @@ set(CLP_FFI_JS_SRC_ZSTD
...
+        ${CLP_FFI_JS_COMMON_LINK_OPTIONS}
+        --emit-tsd=${CLP_FFI_JS_BIN_NAME}.d.ts
+        -sENVIRONMENT=${env}
+    )
+    message(
+        "CLP_FFI_JS_BIN_NAME=\"${CLP_FFI_JS_BIN_NAME}\". \
+CMAKE_BUILD_TYPE=\"${CMAKE_BUILD_TYPE}\". \
+Compile options: ${CLP_FFI_JS_COMMON_COMPILE_OPTIONS}. \
+Link options: ${CLP_FFI_JS_LINK_OPTIONS}"
+    )
```
kirkrodrigues, line 138:
```suggestion
    message(
        "CLP_FFI_JS_BIN_NAME=\"${CLP_FFI_JS_BIN_NAME}\". \
CMAKE_BUILD_TYPE=\"${CMAKE_BUILD_TYPE}\". \
Compile options: ${CLP_FFI_JS_COMMON_COMPILE_OPTIONS}. \
Link options: ${CLP_FFI_JS_LINK_OPTIONS}."
    )
```

Let's also move this log after we call `target_link_options`.

kirkrodrigues (changes requested):

Taskfile.yml
junhaoliao: I agree. Let's see if the new approach makes more sense: 6df29f396f2930ea45d80b81dcf40fb9c684f3d3

Taskfile.yml
```diff
@@ -115,10 +120,12 @@ tasks:
...
-      - >-
-        rsync -a
-        "{{.G_CLP_FFI_JS_BUILD_DIR}}/{{.G_CLP_FFI_JS_TARGET_NAME}}."{d.ts,js,wasm}
-        "{{.OUTPUT_DIR}}/"
+      - for:
+          var: "G_CLP_FFI_JS_ENV_NAMES"
+        cmd: >-
+          rsync -a
+          "{{.G_CLP_FFI_JS_BUILD_DIR}}/{{.G_CLP_FFI_JS_TARGET_PREFIX}}-{{.ITEM}}."{d.ts,js,wasm}
+          "{{.OUTPUT_DIR}}/"
```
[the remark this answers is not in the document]
junhaoliao: thanks. i have fixed it in the latest commit

kirkrodrigues (approved):
Minor nit, but otherwise lgtm.

kirkrodrigues (approved):

PR #1, opened 2024-07-04
Add initial implementation with support for decoding CLP IR streams.

xx12345798:
# Description

1. Integrate ClpIrV1Decoder.cpp authored by @junhaoliao and make necessary modifications.
2. Set up submodules at specified versions according to https://github.com/y-scope/clp/blob/main/components/core/tools/scripts/lib_install/ubuntu-jammy/install-packages-from-source.sh :
   * clp: latest
   * fmt: 8.0.1
   * spdlog: 1.9.2
   * zstd: 1.4.9
3. Add CMakeLists.txt:
   1. Download and extract Boost 1.85.0 source files for `clp`'s header includes.
   2. Specify C++20 standard.
   3. Set the Emscripten executable suffix to .js.
   4. Configure Emscripten compile and link flags for the executable target.
   5. Specify `SPDLOG_FMT_EXTERNAL=1`.
   6. Add target sources and include directories.

# Validation performed

1. Set up Emscirpten SDK as instructed at https://emscripten.org/docs/getting_started/downloads.html . (An automation script is to be PR'ed in the future.)
2. Compiled successfully with these commands:
    ```
    cd <project-root>
    mkdir build
    cd build
    cmake .. -G "Unix Makefiles" -DCMAKE_TOOLCHAIN_FILE=$USER/workspace/emsdk/upstream/emscripten/cmake/Modules/Platform/Emscripten.cmake
    ```

PR #15, opened 2024-10-03
Update to the latest C++ linting configs and process from yscope-dev-utils; Forcibly enable `CMAKE_EXPORT_COMPILE_COMMAND`s.

kirkrodrigues:
# Description

We recently added the C++ linting configs we use across our repos to yscope-dev-utils, as well as a process for using them while allowing repo-specific overrides. Thus, this PR updates to use them and the process. To set them up, this PR adds the task `lint:cpp-configs`.

This PR also forcibly enables `CMAKE_EXPORT_COMPILE_COMMANDS`. Previously, we had written the code such that `CMAKE_EXPORT_COMPILE_COMMANDS` would only be set if it wasn't already set, but for some CMake versions (e.g., 3.16.3), it defaults to `OFF` rather than being unset. Forcibly enabling it should have no known downsides

# Validation performed

* `task lint:cpp-configs` generated symlinks to the C++ linting configs.
  * Running it again succeeded idempotently.
* `task lint:cpp-check` succeeded.
* Added extra spaces in a C++ file and then:
  * `task lint:cpp-check` detected the error.
  * `task lint:cpp-fix` fixed the error.
* Ran clang-tidy manually based on the command in #1 and found no violations.


junhaoliao (approved):
Verified on a Mac ARM computer and observed the `realpath` issue no long exists after git checking out the yscope-dev-utils version in this PR.

For the PR title to be used as the final commit message, how about:
```
Replace C++ linting configs with those managed by yscope-dev-utils; Forcibly enable `CMAKE_EXPORT_COMPILE_COMMANDS`.
```

PR #16, opened 2024-10-04
Pass emsdk path from Taskfile to CMake build; Fix emsdk task name in README.

kirkrodrigues:
# Description

Currently, the `emsdk` path in the Taskfile and the one in CMakeLists.txt (for `CMAKE_TOOLCHAIN_FILE`) are both set to `build/emsdk`; this means changing one requires changing the other. Instead, this PR sets `CMAKE_TOOLCHAIN_FILE` when configuring the CMake project in the Taskfile.

Of course, for IDEs that use CMakeLists.txt directly, they will use the hard-coded path in the CMakeLists.txt which may be out of sync with the one in the Taskfile. Ideally, both the Taskfile and CMakeLists.txt could read from some kind of `.env` file, but that seems like more work than it's worth.

This PR also corrects the `emsdk` task name in the README.

# Validation performed

* `task clp-ffi-js` succeeded.
* Changed `G_EMSDK_DIR` in the Taskfile and `task clp-ffi-js` still succeeded.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Follow the steps below to develop and contribute to the project.
 ## Set up
 Before opening the project in an IDE, you'll first need to download and install [emscripten]:
 ```shell
-task emscripten
+task emsdk
 ```
 
 ## Linting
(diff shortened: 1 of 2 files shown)


junhaoliao (approved):
Validated by changing G_EMSDK_DIR to a directory outside of project root and running `task` still succeeded.

The code changes lgtm and the PR title is fine for the final commit message.

PR #18, opened 2024-10-05
Add tasks to run C++ static analysis using clang-tidy; Separate CMake project config into its own task.

kirkrodrigues:
# Description

This PR adds tasks to run clang-tidy on our C++ code. Since clang-tidy is quite slow, we separate the C++ linting tasks into formatters and static analyzers. The updates to the README describe the new tasks.

We plan to parallelize clang-tidy's checking but it's more work than can be done in this scope of this PR. Similarly (also noted in the code), we currently don't support automated clang-tidy fixes since they can be inaccurate.

# Validation performed

* Renamed `StreamReader::m_ts_pattern` to `StreamReader::mTsPattern` and then:
  * `task lint:check` failed due to the naming violation.
  * `task lint:cpp-check` failed due to the naming violation.
  * `task lint:cpp-static-check` failed due to the naming violation.
  * `task lint:fix` failed due to the naming violation.
  * `task lint:cpp-fix` failed due to the naming violation.
  * `task lint:cpp-static-fix` failed due to the naming violation.
* Added extra whitespace in StreamReader.hpp ad then:
  * `task lint:check` failed due to the violation.
  * `task lint:cpp-check` failed due to the violation.
  * `task lint:cpp-format-check` failed due to the violation.
  * `task lint:fix` failed due to the violation.
  * `task lint:cpp-fix` removed the extra whitespace.
  * `task lint:cpp-format-check` removed the extra whitespace.
* Ran `task lint:cpp-check` twice to ensure no checks were re-run, changed one of the submodules, and then `task lint:cpp-check` detected the changes and re-ran `cpp-static-check`.

## Summary by CodeRabbit

- **New Features**
	- Added detailed linting commands for C++ and YAML in the README.md.
	- Introduced new linting tasks: `cpp-format-check`, `cpp-static-check`, and `clang-tidy` in the linting configuration.
- **Bug Fixes**
	- Enhanced structure and readability of linting tasks configuration.
- **Chores**
	- Updated dependencies to include `clang-tidy` and maintained existing versions for other linting tools.
	- Streamlined the build process with the introduction of the `config-cmake-project` task.

### DIFF
--- a/lint-tasks.yml
+++ b/lint-tasks.yml
@@ -101,6 +101,8 @@ tasks:
         --config-file=.clang-tidy
     deps: [":config-cmake-project", "cpp-configs", "venv"]
     cmd: |-
+      . "{{.G_LINT_VENV_DIR}}/bin/activate"
+
       # Pass emscripten's cflags to clang-tidy by prefixing each one with `--extra-arg`
       EXTRA_ARGS=$("{{.G_EMSDK_DIR}}/upstream/emscripten/em++" --cflags \
         | tr ' ' '\n' \
--- a/lint-tasks.yml
+++ b/lint-tasks.yml
@@ -1,7 +1,8 @@
 version: "3"
 
 vars:
-  G_SRC_CLP_FFI_JS_DIR: "{{.ROOT_DIR}}/src/clp_ffi_js"
+  G_SRC_DIR: "{{.ROOT_DIR}}/src"
+  G_SRC_CLP_FFI_JS_DIR: "{{.G_SRC_DIR}}/clp_ffi_js"
   G_LINT_VENV_DIR: "{{.G_BUILD_DIR}}/lint-venv"
 
 tasks:
@@ -37,6 +38,7 @@ tasks:
       - "{{.G_SRC_CLP_FFI_JS_DIR}}/**/*.hpp"
       - "{{.TASKFILE}}"
       - "tools/yscope-dev-utils/lint-configs/.clang-format"
+    deps: ["cpp-configs", "venv"]
     cmds:
       - task: "clang-format"
         vars:
@@ -44,6 +46,7 @@ tasks:
 
   cpp-format-fix:
     sources: *cpp_format_source_files
+    deps: ["cpp-configs", "venv"]
     cmds:
       - task: "clang-format"
         vars:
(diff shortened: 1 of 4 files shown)


lint-tasks.yml
```diff
@@ -58,14 +85,34 @@ tasks:
...
-          \( -iname "*.cpp" -o -iname "*.h" -o -iname "*.hpp" \) \
-          -print0 | \
-            xargs -0 clang-format {{.FLAGS}} -Werror
+    cmd: |-
+      . "{{.G_LINT_VENV_DIR}}/bin/activate"
+      find "{{.G_SRC_CLP_FFI_JS_DIR}}" \
+        -type f \
+        \( -iname "*.cpp" -o -iname "*.h" -o -iname "*.hpp" \) \
+        -print0 | \
+          xargs -0 clang-format {{.FLAGS}} -Werror
```
kirkrodrigues, line 94: Unrelated to the PR, but this is simple to reformat from `cmds` into ` cmd`.

lint-tasks.yml
```diff
@@ -31,13 +42,29 @@ tasks:
...
         vars:
           FLAGS: "-i"
 
+  cpp-static-check:
+    # Alias task to `cpp-static-fix` since we don't currently support automatic fixes.
+    # NOTE: clang-tidy does have the ability to fix some errors, but the fixes can be inaccurate.
+    # When we eventually determine which errors can be safely fixed, we'll allow clang-tidy to
+    # fix them.
+    aliases: ["cpp-static-fix"]
+    sources:
```
junhaoliao, line 58: Do we need to add `G_EMSDK_CHECKSUM` here?

lint-tasks.yml
```diff
@@ -31,13 +42,29 @@ tasks:
...
+      - "{{.G_SRC_CLP_FFI_JS_DIR}}/**/*.cpp"
+      - "{{.G_SRC_CLP_FFI_JS_DIR}}/**/*.h"
+      - "{{.G_SRC_CLP_FFI_JS_DIR}}/**/*.hpp"
```
junhaoliao, line 61: Do we need to include all source files? Say if some header in `clp-ffi-js`' dependency changes, it would also impact the compilation in `clp-ffi-js`' source code, right?

lint-tasks.yml
kirkrodrigues: Yeah, my mistake. Let me restructure things to be clearer.

lint-tasks.yml
kirkrodrigues: It would, but this task doesn't require compiling clp-ffi-js, right? (I might be misunderstanding your question.)

junhaoliao (dismissed):
As instructed by the validation steps in the PR description, I made a missing header violation in StreamReader.hpp and running each lint task yield expected outputs.
The Taskfile changes meet our guidelines but it seems some extra sources can be added to the `cpp-static-check` sources list to improve accuracy of preventing unnecessary work.
Otherwise everything else lgtm and the PR title is fine to be used as the final commit message.

lint-tasks.yml
junhaoliao: I can't come up with an exact example since I'm not too familiar with clang-tidy rules. However, from my experience working with static analysis, any dependency changes can cause the checks in our code to fail. (In fact, sometime it's easier to fix violations in the dependencies when there are too many violations due to a same issue in the dependency. )

Please correct me if this imaginary example is impossible - say we recently updated some function signature in CLP core code that some parameter changes its type from int64 to int32, but our clp-ffi-js code remains the same. If we keep passing an int64 value to the modified function, wouldn't a precision / value loss violation be raised?

kirkrodrigues: Oh I see what you're saying. Yes, we should include the dependency's source files.

lint-tasks.yml
kirkrodrigues: Since we plan to move clang-format and clang-tidy into yscope-dev-utils, what do you think about the following? Move clang-format and clang-tidy's dependencies into cpp-format-* and cpp-static-* and make clang-format and clang-tidy idempotent.

junhaoliao (approved):
lgtm. The PR title is fine for final commit but we may also want to mention "Add task to streamline CMake project configuration".

kirkrodrigues:
How's the title now?

junhaoliao:
> How's the title now?

Looks good

PR #20, opened 2024-10-10
Update version for 0.2.0 release.

davemarco:
# Description
Update version since last commit had a breaking change for log viewer.

## Summary by CodeRabbit

- **Chores**
	- Updated project version from 0.1.0 to 0.2.0 in `package.json` and `CMakeLists.txt`.
	- Defined Boost library version as 1.85.0.
	- Set executable name for generated JavaScript and WebAssembly files to "ClpFfiJs".
	- Configured C++ standard to C++20 and adjusted build settings.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,7 @@ project(
     LANGUAGES
         C
         CXX
-    VERSION 0.1.0
+    VERSION 0.2.0
 )
 
 # Enable exporting compile commands


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "clp-ffi-js",
-  "version": "0.1.0",
+  "version": "0.2.0",
```
junhaoliao, line 3: @kirkrodrigues I'm not too sure about our conventions - do we need to update the version number in https://github.com/y-scope/clp-ffi-js/blob/424517334bbc81b2f2f30ae8d406b6cff52eafb0/CMakeLists.txt#L17 as well?

junhaoliao (dismissed):
lgtm as updating the "version" in the package.json file should be sufficient for the next release. For the PR title, how about:
```
Update version for 0.2.0 release.
```

package.json
kirkrodrigues: I guess we should, yeah. One day we can hopefully make one depend on the other so we don't have duplicates.

kirkrodrigues (approved):

PR #26, opened 2024-10-31
Split `StreamReader` into an interface and implementation to prepare for adding another IR stream reader.

davemarco:
# Description

Extracts a base class for IRv1 and IRv2 stream readers to inherit from.  Since IRv2 is not yet supported by clp-ffi-js, the repo now includes a base class and a derived class for IRv1.

Code was originally written by Junhao. He made these changes
- Rename ClpIrStreamReader to ClpIRStreamReader in bindings.
- Add get_version to decoding methods.
- Extract a StreamReader base class from IRStreamReader.
- Implement factory method StreamReader::create to perform version checking before creating derived classes.

I merged main afterwards, and made small changes like exporting base class methods instead of derived class.

# Validation performed
Build and test functionality in log viewer (opening a few files and filtering). FYI since the exported class name changed, we will need to update log viewer when this is re-released.

## Summary by CodeRabbit

- **New Features**
	- Introduced the `IrStreamReader` class for reading and processing IR streams.
	- Added methods for filtering log events, deserializing streams, and decoding log event ranges.
	- New function to retrieve version information from a reader interface.
  
- **Bug Fixes**
	- Enhanced error handling during stream deserialization and logging.

- **Refactor**
	- Significant restructuring of the `StreamReader` class to improve memory management and extensibility.

- **Documentation**
	- Updated documentation to reflect changes in class names and functionalities.

## Summary by CodeRabbit

- **New Features**
	- Introduced the `UnstructuredIrStreamReader` class for handling unstructured IR streams, including methods for filtering log events and deserializing data.
	- Enhanced error handling and logging capabilities in the stream processing.

- **Refactor**
	- Major restructuring of the `StreamReader` class, focusing on smart pointer management and interface simplification.
	- Removed several outdated methods to streamline functionality.

- **Chores**
	- Updated build configuration to include new source files and reflect changes in project structure.
	- Updated clang-format configuration to recognize additional library headers.

### DIFF
--- a/src/clp_ffi_js/ir/UnstructuredIrStreamReader.hpp
+++ b/src/clp_ffi_js/ir/UnstructuredIrStreamReader.hpp
@@ -60,6 +60,13 @@ public:
 
     void filter_log_events(LogLevelFilterTsType const& log_level_filter) override;
 
+    /**
+     * @see StreamReader::deserialize_stream
+     *
+     * After the stream has been exhausted, it will be deallocated.
+     *
+     * @return @see StreamReader::deserialize_stream
+     */
     [[nodiscard]] auto deserialize_stream() -> size_t override;
 
     [[nodiscard]] auto decode_range(size_t begin_idx, size_t end_idx, bool use_filter) const


src/clp_ffi_js/ir/IrStreamReader.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+void IrStreamReader::filter_log_events(emscripten::val const& log_level_filter) {
+    if (log_level_filter.isNull()) {
+        m_filtered_log_event_map.reset();
+        return;
+    }
+
+    m_filtered_log_event_map.emplace();
+    auto filter_levels{emscripten::vecFromJSArray<std::underlying_type_t<LogLevel>>(log_level_filter
+    )};
```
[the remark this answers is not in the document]
junhaoliao: I think for now this is skippable since the wrapper code is typed with TypeScript and any usage with non-array would fail static checks.

src/clp_ffi_js/ir/IrStreamReader.cpp
```diff
@@ -0,0 +1,219 @@
...
+        auto result{m_stream_reader_data_context->get_deserializer().deserialize_log_event()};
+        if (result.has_error()) {
+            auto const error{result.error()};
+            if (std::errc::no_message_available == error) {
+                break;
+            }
+            if (std::errc::result_out_of_range == error) {
+                SPDLOG_ERROR("File contains an incomplete IR stream");
+                break;
+            }
```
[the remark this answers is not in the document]
junhaoliao: I think we should not throw an exception here because we intend to handle missing End-of-Stream packet gracefully.

src/clp_ffi_js/ir/IrStreamReader.cpp
```diff
@@ -0,0 +1,264 @@
...
+            log_event_idx = i;
+        }
+        auto const& log_event{m_encoded_log_events[log_event_idx]};
+
+        auto const parsed{log_event.get_message().decode_and_unparse()};
+        if (false == parsed.has_value()) {
+            SPDLOG_ERROR("Failed to decode message.");
+            break;
+        }
+        message = parsed.value();
```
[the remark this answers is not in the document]
junhaoliao: Let's add this. Messages decoding is not expected to fail unless the file has been corrupted / the deserializer in CLP core does not handle edge cases.

src/clp_ffi_js/ir/IrStreamReader.cpp
```diff
@@ -0,0 +1,219 @@
...
```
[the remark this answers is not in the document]
junhaoliao: this seems to be the same as the above comment.

src/clp_ffi_js/ir/IrStreamReader.cpp
```diff
@@ -0,0 +1,219 @@
...
+        EM_ASM(
+                { Emval.toValue($0).push([UTF8ToString($1), $2, $3, $4]); },
+                results.as_handle(),
+                message.c_str(),
+                log_event.get_timestamp(),
+                log_event.get_log_level(),
+                log_event_idx + 1
+        );
+    }
+
```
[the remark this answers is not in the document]
junhaoliao: To be explored in the future.

src/clp_ffi_js/ir/StreamReader.cpp
```diff
@@ -1,44 +1,28 @@
 #include "StreamReader.hpp"
 
-#include <algorithm>
 #include <cstddef>
 #include <cstdint>
-#include <iterator>
+#include <ErrorCode.hpp>
```
junhaoliao, line 5: Not sure how the linter is not catching this but we want to rename this import as 
```
#include <clp/ErrorCode.hpp>`
```
and move it to the import group below.

kirkrodrigues: We misconfigured the include directories in `CMakeLists.txt` It's because we configured `src/submodules/clp/components/core/src/clp` as an include directory in `CMakeLists.txt`, primarily because we treat `string_utils` as an independent library in clp-core, but it's still under the `clp` dir. I'll open an issue in clp to track and fix this. For now, let's double-check that all clp includes aren't mistakenly included in the standard C++ includes.

kirkrodrigues: https://github.com/y-scope/clp/issues/571

src/clp_ffi_js/ir/StreamReaderDataContext.hpp
```diff
@@ -11,7 +11,7 @@
 
 namespace clp_ffi_js::ir {
 /**
- * The data context for a `StreamReader`. It encapsulates a chain of the following resources:
+ * The data context for a `IrStreamReader`. It encapsulates a chain of the following resources:
```
junhaoliao, line 14: We probably don't need change this

src/clp_ffi_js/ir/StreamReader.cpp
```diff
@@ -48,204 +32,36 @@ auto StreamReader::create(DataArrayTsType const& data_array) -> StreamReader {
...
     emscripten::register_type<clp_ffi_js::ir::DecodedResultsTsType>(
             "Array<[string, number, number, number]>"
     );
     emscripten::register_type<clp_ffi_js::ir::FilteredLogEventMapTsType>("number[] | null");
-
-    emscripten::class_<clp_ffi_js::ir::StreamReader>("ClpIrStreamReader")
+    emscripten::class_<clp_ffi_js::ir::StreamReader>("ClpStreamReader")
             .constructor(
                     &clp_ffi_js::ir::StreamReader::create,
                     emscripten::return_value_policy::take_ownership()
```
junhaoliao, line 67: Can we split the `ClpIrStreamReader`'s related bindings into `ClpIrStreamReader.cpp` ?

src/clp_ffi_js/ir/StreamReader.cpp
```diff
@@ -48,204 +32,36 @@ auto StreamReader::create(DataArrayTsType const& data_array) -> StreamReader {
...
-                { Emval.toValue($0).push([UTF8ToString($1), $2, $3, $4]); },
-                results.as_handle(),
-                message.c_str(),
-                log_event.get_timestamp(),
-                log_event.get_log_level(),
-                log_event_idx + 1
+        return std::unique_ptr<IrStreamReader>(
+                new IrStreamReader(std::move(stream_reader_data_context))
         );
     }
```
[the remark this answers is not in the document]
junhaoliao: Given an IRv1 version validate util function is not to be added to CLP-core soon, I agree with this approach.

src/clp_ffi_js/ir/StreamReader.cpp
```diff
@@ -48,204 +32,36 @@ auto StreamReader::create(DataArrayTsType const& data_array) -> StreamReader {
...
+    SPDLOG_CRITICAL("Unable to create stream reader for IR data with version {}.", version);
```
junhaoliao, line 46:
```suggestion
    SPDLOG_CRITICAL("Unable to create stream reader for CLP stream with version {}.", version);
```

src/clp_ffi_js/ir/decoding_methods.cpp
```diff
@@ -33,4 +41,41 @@ auto rewind_reader_and_validate_encoding_type(clp::ReaderInterface& reader) -> v
...
+            metadata_bytes.size()
+    };
+    nlohmann::json const metadata = nlohmann::json::parse(metadata_view);
+
+    // Retrieve version from metadata.
+    auto const& version{metadata.at(clp::ffi::ir_stream::cProtocol::Metadata::VersionKey)};
+    SPDLOG_INFO("The version is {}", version);
+
+    return version;
+}
```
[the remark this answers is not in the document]
junhaoliao: I agree with the rabbit on all those 3 items thought the error messages in the log and throwable should be consistent (don't apply the suggested code directly).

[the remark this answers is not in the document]
junhaoliao: Correction: it is a bad idea to return a string_view in this method since `metadata` goes out of scope when this function returns.

src/clp_ffi_js/ir/StreamReader.cpp
junhaoliao: Sorry I thought we were still to expose the derived classes. If polymorphism / dynamic dispatch works fine, I think it's fine to only creating bindings for `StreamReader`.

src/clp_ffi_js/ir/IrStreamReader.cpp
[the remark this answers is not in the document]
junhaoliao: Actually sorry for my oversight, I think we want to define a type (similar to `FilteredLogEventMapTsType`) instead of directly using `emscripten::val` here.

src/clp_ffi_js/ir/decoding_methods.cpp
```diff
@@ -33,4 +42,48 @@ auto rewind_reader_and_validate_encoding_type(clp::ReaderInterface& reader) -> v
...
+    };
+    if (clp::ffi::ir_stream::IRErrorCode::IRErrorCode_Success != deserialize_preamble_result) {
+        SPDLOG_CRITICAL("Failed to deserialize preamble for version reading");
+
+        throw ClpFfiJsException{
+                clp::ErrorCode::ErrorCode_Failure,
+                __FILENAME__,
+                __LINE__,
+                "Failed to deserialize preamble for version reading"
+        };
```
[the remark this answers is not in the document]
junhaoliao: I agree with this. Let's include the error code in both the log message and the throwable.

src/clp_ffi_js/ir/decoding_methods.hpp
```diff
@@ -1,9 +1,20 @@
 #ifndef CLP_FFI_JS_IR_DECODING_METHODS_HPP
 #define CLP_FFI_JS_IR_DECODING_METHODS_HPP
 
+#include <string>
+
 #include <clp/ReaderInterface.hpp>
 
 namespace clp_ffi_js::ir {
+/**
+ * Gets the version of the IR stream.
```
junhaoliao, line 10:
```suggestion
 * Gets the version of the IR stream from the specified reader.
```

junhaoliao (dismissed):
I haven't tested this but overall it looks good to me. With the remaining comments from the AI bot addressed, we can merge it

src/clp_ffi_js/ir/StreamReader.hpp
```diff
@@ -1,51 +1,40 @@
...
-
 namespace clp_ffi_js::ir {
-using clp::ir::four_byte_encoded_variable_t;
 
 EMSCRIPTEN_DECLARE_VAL_TYPE(DataArrayTsType);
 EMSCRIPTEN_DECLARE_VAL_TYPE(DecodedResultsTsType);
 EMSCRIPTEN_DECLARE_VAL_TYPE(FilteredLogEventMapTsType);
+EMSCRIPTEN_DECLARE_VAL_TYPE(LogLevelFilterTsType);
+
+constexpr std::array<std::string_view, 4> cIrV1Versions = {"v0.0.1", "v0.0.0", "0.0.1", "0.0.0"};
```
junhaoliao, line 18: Let's also add 0.0.2 which added support for UTC offset shift.
```suggestion
constexpr std::array<std::string_view, 6> cIrV1Versions
        = {"v0.0.2", "v0.0.1", "v0.0.0", "0.0.2", "0.0.1", "0.0.0"};
```

junhaoliao, line 18: IRv1 is not a formal name of the text IR stream protocol. Shall we name this `cTextIrVersions`? (if we're to add `cKvPairsIrVersions`.)

cc @kirkrodrigues

src/clp_ffi_js/ir/IrStreamReader.hpp
```diff
@@ -0,0 +1,113 @@
...
+    ~IrStreamReader() override = default;
+
+    // Disable copy constructor and assignment operator
+    IrStreamReader(IrStreamReader const&) = delete;
+    auto operator=(IrStreamReader const&) -> IrStreamReader& = delete;
+
+    // Define default move constructor
+    IrStreamReader(IrStreamReader&&) = default;
+    // Delete move assignment operator since it's also disabled in `clp::ir::LogEventDeserializer`.
+    auto operator=(IrStreamReader&&) -> IrStreamReader& = delete;
```
junhaoliao, line 46: There were some CLP code issues which prevented us from deleting the move constructor with LLVM toolchains (with GCC it could be done). I assume the code issues have been now addressed.
```suggestion
    // Disable move constructor and assignment operator
    IrStreamReader(IrStreamReader&&) = delete;
    auto operator=(IrStreamReader&&) -> IrStreamReader& = delete;
```

src/clp_ffi_js/ir/StreamReader.cpp
```diff
@@ -48,204 +33,40 @@ auto StreamReader::create(DataArrayTsType const& data_array) -> StreamReader {
...
-) -> StreamReaderDataContext<four_byte_encoded_variable_t> {
-    rewind_reader_and_validate_encoding_type(*zstd_decompressor);
-
-    auto result{
-            clp::ir::LogEventDeserializer<four_byte_encoded_variable_t>::create(*zstd_decompressor)
+    throw ClpFfiJsException{
+            clp::ErrorCode::ErrorCode_Unsupported,
+            __FILENAME__,
+            __LINE__,
+            std::format("Unable to create stream reader for IR data with version {}.", version)
```
junhaoliao, line 54:
```suggestion
            std::format("Unable to create reader for CLP stream with version {}.", version)
```

src/clp_ffi_js/ir/StreamReader.hpp
```diff
@@ -1,51 +1,40 @@
...
-    [[nodiscard]] static auto create(DataArrayTsType const& data_array) -> StreamReader;
+    [[nodiscard]] static auto create(DataArrayTsType const& data_array
+    ) -> std::unique_ptr<StreamReader>;
 
     // Destructor
-    ~StreamReader() = default;
+    virtual ~StreamReader() = default;
 
     // Disable copy constructor and assignment operator
     StreamReader(StreamReader const&) = delete;
```
junhaoliao, line 40: Similarly, let delete the move constructor below:
```
    // Disable move constructor and assignment operator
    StreamReader(StreamReader&&) = delete;
    auto operator=(StreamReader&&) -> StreamReader& = delete;
```

src/clp_ffi_js/ir/decoding_methods.hpp
```diff
@@ -1,9 +1,20 @@
...
+/**
+ * Gets the version of the IR stream from the specified reader.
+ * @param reader
+ * @throws ClpFfiJsException if the encoding type couldn't be decoded or the encoding type is
+ * unsupported.
+ * @throws ClpFfiJsException if the preamble could not be deserialized.
+ * @return Stream version.
+ */
+auto get_version(clp::ReaderInterface& reader) -> std::string;
 /**
```
kirkrodrigues, line 18:
```suggestion

/**
```

src/clp_ffi_js/ir/decoding_methods.cpp
```diff
@@ -33,4 +42,54 @@ auto rewind_reader_and_validate_encoding_type(clp::ReaderInterface& reader) -> v
         };
     }
 }
+
+auto get_version(clp::ReaderInterface& reader) -> std::string {
```
kirkrodrigues, line 46: Style violation: Functions/method definitions should be ordered in the same order as their declarations.

src/clp_ffi_js/ir/StreamReader.hpp
kirkrodrigues: Sgtm

src/clp_ffi_js/ir/IrStreamReader.hpp
kirkrodrigues: Not yet, unfortunately. I have the fix on a branch, but it's part of a larger change that needs to be split up and was waiting for some other PRs to get merged. It should end up being merged sometime this week.

src/clp_ffi_js/ir/IrStreamReader.hpp
```diff
@@ -0,0 +1,113 @@
+#ifndef CLP_FFI_JS_IR_IR_STREAM_READER_HPP
```
kirkrodrigues, line 1: I'll add a linter for this soon, but the format of header defines should be `<NAMESPACE>_<FILENAME>_HPP`, which in this case would be `CLP_FFI_JS_IR_IRSTREAMREADER_HPP`.

src/clp_ffi_js/ir/IrStreamReader.hpp
```diff
@@ -0,0 +1,113 @@
...
+
+/**
+ * Mapping between an index in the filtered log events collection to an index in the unfiltered
+ * log events collection.
+ */
+using FilteredLogEventsMap = std::optional<std::vector<size_t>>;
+
+/**
+ * Class to deserialize and decode Zstandard-compressed CLP IRv1 streams as well as format decoded
+ * log events.
```
kirkrodrigues, line 30:
```suggestion
 * Class to deserialize and decode Zstandard-compressed CLP text IR streams, as well as format
 * decoded log events.
```

src/clp_ffi_js/ir/IrStreamReader.hpp
```diff
@@ -0,0 +1,113 @@
...
+
+    // Define default move constructor
+    IrStreamReader(IrStreamReader&&) = default;
+    // Delete move assignment operator since it's also disabled in `clp::ir::LogEventDeserializer`.
+    auto operator=(IrStreamReader&&) -> IrStreamReader& = delete;
+
+    // Methods
+    /**
+     * @return The number of events buffered.
+     */
```
kirkrodrigues, line 51:
```suggestion
```
We don't need to repeat the docstring for implementations of interface methods unless the semantics are different than those described in the interface (added this point to the internal coding guidelines).

src/clp_ffi_js/ir/IrStreamReader.hpp
```diff
@@ -0,0 +1,113 @@
...
+    [[nodiscard]] auto get_num_events_buffered() const -> size_t override;
+
+    /**
+     * @return The filtered log events map.
+     */
```
kirkrodrigues, line 56:
```suggestion
```

src/clp_ffi_js/ir/IrStreamReader.hpp
```diff
@@ -0,0 +1,113 @@
...
+    [[nodiscard]] auto get_filtered_log_event_map() const -> FilteredLogEventMapTsType override;
+
+    /**
+     * Generates a filtered collection from all log events.
+     *
+     * @param log_level_filter Array of selected log levels
+     */
```
kirkrodrigues, line 63:
```suggestion
```

src/clp_ffi_js/ir/IrStreamReader.hpp
```diff
@@ -0,0 +1,113 @@
...
+    void filter_log_events(LogLevelFilterTsType const& log_level_filter) override;
+
+    /**
+     * Deserializes all log events in the stream. After the stream has been exhausted, it will be
```
kirkrodrigues, line 67: Deallocation seems like an implementation decision, so perhaps we can keep this note in the implementation's docstring but not in the interface's. @junhaoliao, what do you think?

src/clp_ffi_js/ir/IrStreamReader.hpp
```diff
@@ -0,0 +1,113 @@
...
+     * @param end_idx
+     * @param use_filter Whether to decode from the filtered or unfiltered log events collection.
+     * @return An array where each element is a decoded log event represented by an array of:
+     * - The log event's message
+     * - The log event's timestamp as milliseconds since the Unix epoch
+     * - The log event's log level as an integer that indexes into `cLogLevelNames`
+     * - The log event's number (1-indexed) in the stream
+     * @return null if any log event in the range doesn't exist (e.g. the range exceeds the number
+     * of log events in the collection).
+     */
```
kirkrodrigues, line 88:
```suggestion
```

src/clp_ffi_js/ir/IrStreamReader.hpp
```diff
@@ -0,0 +1,113 @@
...
+    [[nodiscard]] auto decode_range(size_t begin_idx, size_t end_idx, bool use_filter) const
+            -> DecodedResultsTsType override;
+
+private:
+    // Constructor
+    IrStreamReader(
```
kirkrodrigues, line 94:
```suggestion
    explicit IrStreamReader(
```

Unless Junhao was the one who suggested removing it?

src/clp_ffi_js/ir/IrStreamReader.cpp
```diff
@@ -0,0 +1,224 @@
...
+            }
+            if (std::errc::result_out_of_range == error) {
+                SPDLOG_ERROR("File contains an incomplete IR stream");
+                break;
+            }
+            throw ClpFfiJsException{
+                    clp::ErrorCode::ErrorCode_Corrupt,
+                    __FILENAME__,
+                    __LINE__,
+                    "Failed to deserialize: "s + error.category().name() + ":" + error.message()
```
kirkrodrigues, line 96: Use `std::format`?

src/clp_ffi_js/ir/IrStreamReader.cpp
```diff
@@ -0,0 +1,224 @@
...
+        if (use_filter) {
+            log_event_idx = m_filtered_log_event_map->at(i);
+        } else {
+            log_event_idx = i;
+        }
+        auto const& log_event{m_encoded_log_events[log_event_idx]};
+
+        auto const parsed{log_event.get_message().decode_and_unparse()};
+        if (false == parsed.has_value()) {
+            SPDLOG_ERROR("Failed to decode message.");
```
kirkrodrigues, line 165: Necessary since we have the exception?

kirkrodrigues, line 165: Necessary since we have the exception?

src/clp_ffi_js/ir/IrStreamReader.cpp
```diff
@@ -0,0 +1,224 @@
...
+    auto result{
+            clp::ir::LogEventDeserializer<four_byte_encoded_variable_t>::create(*zstd_decompressor)
+    };
+    if (result.has_error()) {
+        auto const error_code{result.error()};
+        SPDLOG_CRITICAL(
+                "Failed to create deserializer: {}:{}",
+                error_code.category().name(),
+                error_code.message()
+        );
```
kirkrodrigues, line 213: Merge this log into the the exception?

src/clp_ffi_js/ir/StreamReader.cpp
```diff
@@ -48,204 +33,40 @@ auto StreamReader::create(DataArrayTsType const& data_array) -> StreamReader {
...
-                results.as_handle(),
-                message.c_str(),
-                log_event.get_timestamp(),
-                log_event.get_log_level(),
-                log_event_idx + 1
+        return std::unique_ptr<IrStreamReader>(
+                new IrStreamReader(std::move(stream_reader_data_context))
         );
     }
+    SPDLOG_CRITICAL("Unable to create reader for CLP stream with version {}.", version);
```
kirkrodrigues, line 48:
```suggestion
```

Since I think it's covered by the exception.

src/clp_ffi_js/ir/StreamReader.cpp
```diff
@@ -48,204 +33,40 @@ auto StreamReader::create(DataArrayTsType const& data_array) -> StreamReader {
...
-) -> StreamReaderDataContext<four_byte_encoded_variable_t> {
-    rewind_reader_and_validate_encoding_type(*zstd_decompressor);
-
-    auto result{
-            clp::ir::LogEventDeserializer<four_byte_encoded_variable_t>::create(*zstd_decompressor)
+    throw ClpFfiJsException{
+            clp::ErrorCode::ErrorCode_Unsupported,
+            __FILENAME__,
+            __LINE__,
+            std::format("Unable to create reader for CLP stream with version {}.", version)
```
kirkrodrigues, line 54:
```suggestion
            std::format("Unable to create reader for IR stream with version {}.", version)
```

src/clp_ffi_js/ir/IrStreamReader.hpp
```diff
@@ -0,0 +1,113 @@
...
+ * Mapping between an index in the filtered log events collection to an index in the unfiltered
+ * log events collection.
+ */
+using FilteredLogEventsMap = std::optional<std::vector<size_t>>;
+
+/**
+ * Class to deserialize and decode Zstandard-compressed CLP IRv1 streams as well as format decoded
+ * log events.
+ */
+class IrStreamReader : public StreamReader {
```
kirkrodrigues, line 32: How about calling this `TextIrStreamReader` and `StreamReader` -> `IrStreamReader`?

src/clp_ffi_js/ir/StreamReader.cpp
```diff
@@ -48,204 +33,40 @@ auto StreamReader::create(DataArrayTsType const& data_array) -> StreamReader {
...
-    }
-
-    return {std::move(data_buffer), std::move(zstd_decompressor), std::move(result.value())};
 }
 }  // namespace clp_ffi_js::ir
 
 namespace {
-EMSCRIPTEN_BINDINGS(ClpIrStreamReader) {
+EMSCRIPTEN_BINDINGS(ClpStreamReader) {
+    // Output to JS types
```
kirkrodrigues, line 61:
```suggestion
    // JS types used as inputs
```

src/clp_ffi_js/ir/StreamReader.cpp
```diff
@@ -48,204 +33,40 @@ auto StreamReader::create(DataArrayTsType const& data_array) -> StreamReader {
...
     emscripten::register_type<clp_ffi_js::ir::DataArrayTsType>("Uint8Array");
     emscripten::register_type<clp_ffi_js::ir::DecodedResultsTsType>(
             "Array<[string, number, number, number]>"
     );
     emscripten::register_type<clp_ffi_js::ir::FilteredLogEventMapTsType>("number[] | null");
-
-    emscripten::class_<clp_ffi_js::ir::StreamReader>("ClpIrStreamReader")
+    // Input from JS types
```
kirkrodrigues, line 67:
```suggestion

    // JS types used as outputs
```

One thing I've picked up from Junhao's eslint rules is ensuring there's a newline before comments.

src/clp_ffi_js/ir/StreamReader.cpp
```diff
@@ -48,204 +33,40 @@ auto StreamReader::create(DataArrayTsType const& data_array) -> StreamReader {
...
+    auto const version{get_version(*zstd_decompressor)};
 
-        auto log_viewer_event{LogEventWithLevel<four_byte_encoded_variable_t>(
-                log_event.get_timestamp(),
-                log_event.get_utc_offset(),
-                message,
-                log_level
+    if (std::ranges::find(cTextIrVersions, version) != cTextIrVersions.end()) {
+        auto stream_reader_data_context{IrStreamReader::create_data_context(
+                std::move(zstd_decompressor),
```
kirkrodrigues, line 40: I think it's strange for StreamReader to be calling this method:

* The method is `IrStreamReader`-specific since it deals with the `LogEventDeserializer` class which, although it's name doesn't suggest it, is a text-IR-specific class.
* The method is private, so we have to make the base class access a private method of the derived class through the friend access path.

Instead, can we:
* Create `IrStreamReader::create` that handles constructing the `StreamReaderContext` and the `IrStreamReader`.
* Rename `StreamReaderContext` to `IrStreamReaderContext`.
* Replace the logic here with a call to `IrStreamReader::create`.

kirkrodrigues (changes requested):
Review of remainder.

src/clp_ffi_js/ir/decoding_methods.cpp
```diff
@@ -33,4 +42,54 @@ auto rewind_reader_and_validate_encoding_type(clp::ReaderInterface& reader) -> v
...
+    // The encoding type bytes must be consumed before the metadata can be read.
+    rewind_reader_and_validate_encoding_type(reader);
```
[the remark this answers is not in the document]
kirkrodrigues: > Note if take the call to rewind_reader_and_validate_encoding_type out of get_version is it not safe to call by itself.

That is true, so in some sense, I think `get_version` should be a method internal to `StreamReader.cpp`. I don't think it's worth it to make it safe to use outside that context (by e.g., validating that the stream position is correct for it to be called). Overall, I think the semantic we want to maintain is that the base class reads the encoding type and the derived classes read the rest.

src/clp_ffi_js/ir/StreamReader.cpp
kirkrodrigues: Sorry, I didn't mean to switch them. Are you okay with the suggestion if we don't switch input with output?

src/clp_ffi_js/ir/IrStreamReader.hpp
kirkrodrigues: How about inserting `Ir` before `Stream`---`UnstructuredIrStreamReader` and `StructuredIrStreamReader`? (Once the new IR format supports backward compatibility, hopefully we can revert back to a single `IrStreamReader` class.)

> since he wants to one day have an ffi for single file archive.

I see. I guess we can keep `StreamReader` then.

> Text seems a bit vague

It's how we differentiate unstructured and structured for the casual user. "Text" for unstructured and "JSON" for structured.

src/clp_ffi_js/ir/IrStreamReader.hpp
kirkrodrigues: Hm, looking at the history, I think he made that comment about a constructor that takes no arguments, right?

src/clp_ffi_js/ir/StreamReader.cpp
junhaoliao: Sorry for missing this earlier. 

> Rename StreamReaderContext to IrStreamReaderContext.
I thought we could templatize class `StreamReaderDataContext` to encapsulate either a text-IR or kvPairs-IR deserializer, which avoids creating a separate data context class for kvPairs-IR. 

I agree with the other two changes.

src/clp_ffi_js/ir/IrStreamReader.hpp
junhaoliao: Sure

src/clp_ffi_js/ir/StreamReader.cpp
junhaoliao: > junhao prefers no IR here

Sorry! Pretty sure it was me who made a suggestion in this line but the suggestion was wrong. Using `IR` would be more clear in this context since the stream reader is IR specific.

Just so we don't confuse ourselves and to make sure we're on the same page, there're cases we want to avoid certain terms:
1. IRv1 / IRv2: instead UnstructuredIR / KvPairsIR should be used
2. StreamReaderDataContext: I was trying to avoid mentioning IR at all since the class once templated with deserializer type, it can be used for Single Archive CLP as well. (That said, we have a pending discussion at https://github.com/y-scope/clp-ffi-js/pull/26#discussion_r1830665669

src/clp_ffi_js/ir/StreamReader.cpp
kirkrodrigues: Would you mind renaming it now and then templating it once we add the KV-pair IR stream reader?

src/clp_ffi_js/ir/StreamReader.cpp
kirkrodrigues: 1. How about UnstructuredIr and StructuredIr? That's what we were thinking in https://github.com/y-scope/clp-ffi-js/pull/26#discussion_r1830241353 I know it's a departure from what we're using in clp-core, but clp-core's naming convention is also confusing by itself since the UnstructuredIr format doesn't have a special name. And hopefully once we support backwards compatibility in the new IR format, we won't have any of these issues any more.

junhaoliao: Sounds good

src/clp_ffi_js/ir/StreamReader.cpp
junhaoliao: I see both your points. At the moment, I feel it might not be too bad to leave the name given this class currently resides in path src/clp-ffi-js/ir

kirkrodrigues: Sure

src/clp_ffi_js/ir/StreamReader.hpp
```diff
@@ -1,51 +1,47 @@
...
-#include <clp/TimestampPattern.hpp>
-#include <emscripten/bind.h>
+#include <clp/streaming_compression/zstd/Decompressor.hpp>
 #include <emscripten/val.h>
 
-#include <clp_ffi_js/ir/LogEventWithLevel.hpp>
-#include <clp_ffi_js/ir/StreamReaderDataContext.hpp>
-
 namespace clp_ffi_js::ir {
-using clp::ir::four_byte_encoded_variable_t;
```
kirkrodrigues, line 13:
```suggestion
```

src/clp_ffi_js/ir/StreamReader.hpp
```diff
@@ -57,30 +53,30 @@ class StreamReader {
     auto operator=(StreamReader&&) -> StreamReader& = delete;
 
     // Methods
+
```
kirkrodrigues, line 56:
```suggestion
```

src/clp_ffi_js/ir/StreamReader.hpp
```diff
@@ -97,28 +93,11 @@ class StreamReader {
...
-            m_stream_reader_data_context;
-    FilteredLogEventsMap m_filtered_log_event_map;
-    clp::TimestampPattern m_ts_pattern;
+protected:
+    explicit StreamReader() = default;
 };
 }  // namespace clp_ffi_js::ir
-
-#endif  // CLP_FFI_JS_IR_STREAM_READER_HPP
+#endif  // CLP_FFI_JS_IR_STREAMREADER_HPP
```
kirkrodrigues, line 103:
```suggestion

#endif  // CLP_FFI_JS_IR_STREAMREADER_HPP
```

src/clp_ffi_js/ir/UnstructuredIrStreamReader.hpp
```diff
@@ -0,0 +1,85 @@
...
+    UnstructuredIrStreamReader(UnstructuredIrStreamReader&&) = default;
+    // Delete move assignment operator since it's also disabled in `clp::ir::LogEventDeserializer`.
+    auto operator=(UnstructuredIrStreamReader&&) -> UnstructuredIrStreamReader& = delete;
+
+    /**
+     * First packages a `StreamReaderDataContext` using inputs and a unstructured IR deserializer,
+     * then creates a `UnstructuredIrStreamReader`.
+     *
+     * @param zstd_decompressor
+     * @param data_array An array containing a Zstandard-compressed IR stream.
```
kirkrodrigues, line 50:
```suggestion
     * @param zstd_decompressor A decompressor for an IR stream, where the read head of the stream
     * is just after the stream's encoding type.
     * @param data_array The array backing `zstd_decompressor`.
```

I deleted the description because I think it's mostly implementation details. I think the important parts for the caller are what I added in the parameter descriptions.

src/clp_ffi_js/ir/UnstructuredIrStreamReader.hpp
```diff
@@ -0,0 +1,85 @@
...
+     * @return The created instance.
+     * @throw ClpFfiJsException if any error occurs.
+     */
+    [[nodiscard]] static auto create(
+            std::unique_ptr<ZstdDecompressor>&& zstd_decompressor,
+            clp::Array<char>&& data_array
```
kirkrodrigues, line 56: Let's pass this as a value for the same reason I discussed with Junhao in a previous PR. Essentially, passing it as a value and moving it in the caller offers maximum flexibility. We can't do the same for zstd_decompressor because the class doesn't support it and likely won't any time soon.

src/clp_ffi_js/ir/IrStreamReader.hpp
kirkrodrigues: Currently, I see you've deleted this docstring. I'm fine keeping it this way, but if @junhaoliao wants to keep the part about deallocating the stream after exhaustion, then we need to add the docstring back in `UnstructuredIrStreamReader`.

src/clp_ffi_js/ir/StreamReader.cpp
```diff
@@ -3,249 +3,122 @@
...
+using ClpFfiJsException = clp_ffi_js::ClpFfiJsException;
+using IRErrorCode = clp::ffi::ir_stream::IRErrorCode;
+
+/**
+ * Rewinds the reader to start then validates the CLP IR data encoding type.
+ * @param reader
+ * @throws ClpFfiJsException if the encoding type couldn't be decoded or the encoding type is
+ * unsupported.
+ */
+auto rewind_reader_and_validate_encoding_type(clp::ReaderInterface& reader) -> void {
```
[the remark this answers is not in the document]
kirkrodrigues: Yes, it should be at the top of the anonymous namespace, i.e.:
* Function definitions.
* Function declarations.

src/clp_ffi_js/.clang-format
```diff
@@ -7,7 +7,7 @@ IncludeCategories:
     Priority: 4
   # Library headers. Update when adding new libraries.
   # NOTE: clang-format retains leading white-space on a line in violation of the YAML spec.
-  - Regex: "<(emscripten|fmt|spdlog)"
+  - Regex: "<(emscripten|fmt|spdlog|json)"
```
kirkrodrigues, line 10:
```suggestion
  - Regex: "<(emscripten|fmt|json|spdlog)"
```

src/clp_ffi_js/ir/StreamReader.cpp
```diff
@@ -263,3 +141,46 @@ EMSCRIPTEN_BINDINGS(ClpIrStreamReader) {
...
+    // NOLINTBEGIN(cppcoreguidelines-pro-type-reinterpret-cast)
+    emscripten::val::module_property("HEAPU8")
+            .call<void>("set", data_array, reinterpret_cast<uintptr_t>(data_buffer.data()));
+    // NOLINTEND(cppcoreguidelines-pro-type-reinterpret-cast)
+
+    auto zstd_decompressor{std::make_unique<ZstdDecompressor>()};
+    zstd_decompressor->open(data_buffer.data(), length);
+
+    rewind_reader_and_validate_encoding_type(*zstd_decompressor);
+    // Validate the stream's version
```
kirkrodrigues, line 161:
```suggestion

    // Validate the stream's version
```

src/clp_ffi_js/ir/StreamReader.cpp
```diff
@@ -263,3 +141,46 @@ EMSCRIPTEN_BINDINGS(ClpIrStreamReader) {
...
+        zstd_decompressor->seek_from_begin(pos);
+    } catch (ZstdDecompressor::OperationFailed& e) {
+        throw ClpFfiJsException{
+                clp::ErrorCode::ErrorCode_Failure,
+                __FILENAME__,
+                __LINE__,
+                std::format("Unable to rewind zstd decompressor: {}", e.what())
+        };
+    }
+    return std::make_unique<UnstructuredIrStreamReader>(
```
kirkrodrigues, line 182:
```suggestion

    return std::make_unique<UnstructuredIrStreamReader>(
```

kirkrodrigues (changes requested):

src/clp_ffi_js/ir/IrStreamReader.hpp
junhaoliao: (just to be sure I understand the discussion - 
* in `StreamReader` which serves an interface, I agree we do not want to mention `After the stream has been exhausted, it will be deallocated.`.
* in `UnstructuredIrStreamReader` which is an implementation of the interface, I prefer that we mention `After the stream has been exhausted, it will be deallocated.` so that we can remind any future developer about the stream data lifecycle.

kirkrodrigues: Correct. One way to do this is:

```cpp
    /**
     * Deserializes all log events in the stream. After the stream has been exhausted, it will be
     * deallocated.
     *
     * @return The number of successfully deserialized ("valid") log events.
     */
```

Another way is:

```cpp
    /**
     * See `StreamReader::deserialize_stream`.
     * 
     * After the stream has been exhausted, it will be deallocated.
     *
     * @return See `StreamReader::deserialize_stream`.
     */
```

We usually go with the first approach, but we don't have too many instances of this case, so we could change conventions.

junhaoliao: I think we can go with the second approach to avoid repetition and docs going out of sync any time in the future.

For the second approach, can we use `@sa` or `@see`? (i'm not too sure if this is how the command is intended to be used: https://www.doxygen.nl/manual/commands.html#cmdsa

kirkrodrigues: Didn't know Doxygen had see. Yeah, let's use `@see`.

kirkrodrigues: @davemarco, can you make the change?

kirkrodrigues (approved):
For the PR title, how about:

```
Split `StreamReader` into an interface and implementation to prepare for adding another IR stream reader.
```

PR #29, opened 2024-11-07
Add StructuredIrStreamReader.

junhaoliao:
# Description

# Validation performed


src/clp_ffi_js/ir/StructuredIrStreamReader.hpp
```diff
@@ -0,0 +1,169 @@
...
+#include <clp_ffi_js/ir/StreamReader.hpp>
+#include <clp_ffi_js/ir/StreamReaderDataContext.hpp>
+
+namespace clp_ffi_js::ir {
+using parsed_tree_node_id_t = std::optional<clp::ffi::SchemaTree::Node::id_t>;
+
+/**
+ * Class to handle deserialized IR units.
+ */
+class IrUnitHandler {
```
junhaoliao, line 27: Do we want to move this class into a separate header? That way we don't have to write `using StructuredIrDeserializer = clp::ffi::ir_stream::Deserializer<IrUnitHandler>;` after this class.

PR #19, opened 2024-10-05
Allow using other CMake generators besides "Unix Makefiles".

junhaoliao:
# Description

Previously, it was mandated in CMakeList.txt that only `Unix Makefiles` generators can be used because compilation errors were observed with code that references C++-20 headers. It is confirmed that now with Ninja generators, the `clp-ffi-js` code compiles fine with Emsdk @ 3.1.67 .

1. Remove "Unix Makefiles" `CMAKE_GENERATOR` check.
2. Eliminate the generator specification from compilation commands in Taskfile.

# Validation performed

1. On a Mac computer.
    1. Ran `task clp-ffi-js` and observed compilation success.
    2. Added a semantically meaningless newline at the end of StreamReader.cpp and ran `task clp-ffi-js` again. Observed the source code getting compiled and the overall compilation succeeded.
2. On an Ubuntu 20.04 computer, repeat above validation steps. (In my case, I was using WSL. )

## Summary by CodeRabbit

- **New Features**
	- Simplified build configuration by allowing various generators without restrictions.
	- Introduced new variables in task definitions for improved build clarity and reliability.
- **Bug Fixes**
	- Enhanced handling of the Boost library version for better compatibility.
- **Chores**
	- Updated task commands to ensure a clean build environment and consistent checksum validation.

PR #32, opened 2024-11-09
Add `LogEventWithFilterData` template class to extend log event classes with processed fields for filtering.

davemarco:
# Description
Replace `LogEventWithLevel` with `LogEventWithFilterData`. `LogEventWithFilterData` is similar to `LogEventWithLevel`, however it uses a template instead of inheritance; therefore, it can be used for both unstructured and structured log events. The idea for this class is that it will store the parsed log level and timestamp while the IR file is being deserialized for both unstructured and structured IR streams. 

The log level must be parsed for structured streams since the log level value may be invalid (i.e. the value is not a known log level string).

While we could parse and validate the structured log level during filtering and decoding, it is less repetitive to do it only once when the stream is being deserialized. 

# Validation performed
Tested with modified log viewer. Unstructured still works

## Summary by CodeRabbit

- **New Features**
	- Introduced a new log event class, `LogEventWithFilterData`, enhancing log level filtering capabilities.
	- Added methods to retrieve log event details, including log level and timestamp.

- **Bug Fixes**
	- Updated log event handling in the `UnstructuredIrStreamReader` to ensure compatibility with the new log event structure.

- **Chores**
	- Removed the outdated `LogEventWithLevel` class to streamline the codebase.

### DIFF
--- a/src/clp_ffi_js/ir/LogEventWithFilterData.hpp
+++ b/src/clp_ffi_js/ir/LogEventWithFilterData.hpp
@@ -1,6 +1,7 @@
 #ifndef CLP_FFI_JS_IR_LOGEVENTWITHFILTERDATA_HPP
 #define CLP_FFI_JS_IR_LOGEVENTWITHFILTERDATA_HPP
 
+#include <concepts>
 #include <utility>
 
 #include <clp/ir/LogEvent.hpp>


src/clp_ffi_js/ir/LogEventWithFilterData.hpp
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * A class which accepts a log event type as a template parameter and provides additional members
+ * for the log level and timestamp. The additional members facilitate log level filtering.
+ * @tparam LogEvent The type of the log event.
+ */
+template <typename LogEvent>
+class LogEventWithFilterData {
+public:
+    // Constructor
+    explicit LogEventWithFilterData(
```
junhaoliao, line 20: Since we have more than one parameters in this constructor, `explicit` is not meaningful here.
```suggestion
    LogEventWithFilterData(
```

src/clp_ffi_js/ir/LogEventWithFilterData.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+#include <clp_ffi_js/constants.hpp>
+
+namespace clp_ffi_js::ir {
+/**
+ * A class which accepts a log event type as a template parameter and provides additional members
+ * for the log level and timestamp. The additional members facilitate log level filtering.
+ * @tparam LogEvent The type of the log event.
+ */
+template <typename LogEvent>
```
junhaoliao, line 16: If I'm not wrong, all template parameters should following naming convention `type_name_t`.
cc @LinZhihao-723 / @kirkrodrigues please help confirm.

kirkrodrigues: Only primitive template parameters are named following the format type_name_t. Non-primitives are named how Marco has them.

LinZhihao-723: A little adding on:
Since we're using C++20, we should use `requires` or define a concept about the accepted LogEvent type

src/clp_ffi_js/ir/LogEventWithFilterData.hpp
```diff
@@ -0,0 +1,53 @@
...
+#include <utility>
+
+#include <clp/ir/types.hpp>
+
+#include <clp_ffi_js/constants.hpp>
+
+namespace clp_ffi_js::ir {
+/**
+ * A class which accepts a log event type as a template parameter and provides additional members
+ * for the log level and timestamp. The additional members facilitate log level filtering.
```
junhaoliao, line 13: I understand we are including the timestamp in this class because we want to enable timestamp parsing for the `StructuredStreamReader`; however, the intention of such is not very explicit in this PR.
Shall we say
```
A templated class that accepts a log event type and extends it with additional members for log level and timestamp. These members enable efficient filtering by log level and timestamp, enhancing flexibility in log processing.
```

src/clp_ffi_js/ir/LogEventWithFilterData.hpp
junhaoliao: @davemarco caution, i could be mistaken but i remember having issues with Emscripten + Ninja + C++20 features when we initialize this project, and not finding the `concepts` headers were one of the symptoms. The mitigation to use Unix Makefiles was removed in #19 . You may try adding the mitigation back if you see odd issues compiling the code.

junhaoliao (commented):
lgtm. Since I prototyped part of this code, someone else please help review and approve if appropriate.

src/clp_ffi_js/ir/LogEventWithFilterData.hpp
kirkrodrigues, line 13:
```suggestion
 * A templated class that extends a log event type with processed versions of some of its fields,
 * specifically the fields that are used for filtering in the `StreamReader` classes and their
 * callers.
 *
```

src/clp_ffi_js/ir/LogEventWithFilterData.hpp
```diff
@@ -0,0 +1,53 @@
...
+    LogEventWithFilterData(
+            LogEvent log_event,
+            LogLevel log_level,
+            clp::ir::epoch_time_ms_t timestamp
+    )
+            : m_log_event{std::move(log_event)},
+              m_log_level{log_level},
+              m_timestamp{timestamp} {}
+
+    // Disable copy constructor and assignment operator
```
kirkrodrigues, line 29: Why?

src/clp_ffi_js/ir/LogEventWithFilterData.hpp
kirkrodrigues: Although the types that this class will accept don't need to conform to any interface, the description I suggested does indicate that they need to be log events with fields. Not all types (e.g. `int64_t`) would fit that description (unless we're using a very loose definition), so in that case, using a concept/requires statement to restrict the types to the existing LogEvent types would be beneficial.

Let me know if you disagree.

src/clp_ffi_js/ir/UnstructuredIrStreamReader.cpp
```diff
@@ -147,13 +147,12 @@ auto UnstructuredIrStreamReader::deserialize_stream() -> size_t {
...
-                log_event.get_utc_offset(),
-                message,
-                log_level
+        auto log_event_with_filter_data{LogEventWithFilterData<UnstructuredLogEvent>(
+                log_event,
+                log_level,
+                log_event.get_timestamp()
         )};
-        m_encoded_log_events.emplace_back(std::move(log_viewer_event));
+        m_encoded_log_events.emplace_back(std::move(log_event_with_filter_data));
```
kirkrodrigues, line 155: As the rabbit suggested, we could emplace directly?

src/clp_ffi_js/ir/UnstructuredIrStreamReader.cpp
```diff
@@ -187,9 +186,13 @@ auto UnstructuredIrStreamReader::decode_range(size_t begin_idx, size_t end_idx,
         } else {
             log_event_idx = i;
         }
-        auto const& log_event{m_encoded_log_events[log_event_idx]};
+        auto const& log_event_with_filter_data{m_encoded_log_events[log_event_idx]};
+        auto const& log_level = log_event_with_filter_data.get_log_level();
+        auto const& timestamp = log_event_with_filter_data.get_timestamp();
 
-        auto const parsed{log_event.get_message().decode_and_unparse()};
+        auto const& unstructured_log_event = log_event_with_filter_data.get_log_event();
```
kirkrodrigues, line 193: How about inserting this above line 190 so that we're essentially pulling out all the fields of the `LogEventWithFilterData`?

kirkrodrigues (changes requested):

src/clp_ffi_js/ir/LogEventWithFilterData.hpp
junhaoliao: All good except I think you meant to capitalize `StreamReader`.

junhaoliao: Sorry I just read the latest code. I think that's good enough. (Kirk might have better ideas about improving that though

src/clp_ffi_js/ir/UnstructuredIrStreamReader.cpp
junhaoliao: If I'm not getting Kirk's comment wrong, I think he meant we could get rid of the separated variable declaration and directly write:
```
        m_encoded_log_events.emplace_back(
                log_event,
                log_level,
                log_event.get_timestamp()
        );
```

kirkrodrigues: Yeah, that's what I mean.

src/clp_ffi_js/ir/LogEventWithFilterData.hpp
```diff
@@ -0,0 +1,66 @@
...
+#include <clp_ffi_js/constants.hpp>
+
+namespace clp_ffi_js::ir {
+using clp::ir::four_byte_encoded_variable_t;
+using UnstructuredLogEvent = clp::ir::LogEvent<four_byte_encoded_variable_t>;
+
+// Concept defining valid log event types.
+// TODO: Extend valid log event types when filtering support is added for structured logs.
+template <typename T>
+concept ValidLogEventTypes = std::same_as<T, UnstructuredLogEvent>;
```
kirkrodrigues, line 19: Why not inline the concept into the requires statement? We don't have formal guidelines on using concepts yet, but I feel like we can avoid creating a concept for cases like this.

src/clp_ffi_js/ir/LogEventWithFilterData.hpp
kirkrodrigues: > The key value pair log event has it.

Yeah, I disagree with that decision, lol. I will follow up on that separately.

> So felt it made sense to have it here.

Fair enough. I will come back to this once we resolve the issue with KeyValuePairLogEvent.

> It also warns people if they are trying to copy it when it can be moved.

I think that argument is a bit of a slippery slope unless you have a type that is truly too massive to be copied. Yes, we want users to write high-performance code, but trying to force them to do that by disabling copying just hurts users who really want a copy of an object. That user may then want to add a clone member to your class at which point we're just adding syntax sugar for the sake of preventing code that *might* perform poorly if the compiler doesn't already optimize it.

kirkrodrigues (approved):
For the PR title, how about:

```
Add `LogEventWithFilterData` template class to extend log event classes with processed fields for filtering.
```

PR #35, opened 2024-11-11
feat: Add support for log-level filtering of structured IR streams.

davemarco:
# Description
Adds log level filtering for structured reader.

Design is similiar to junhao prototype. We retrieve log level and timestamp node id from the new node ir unit as the stream is being deserialized. Then we use the cached node id to retrieve values from the logs. 

Difference from prototype is that parsing is done in deserialization stage.

I also added support for both string and int log levels. 

# Validation performed
I tested on one log file with a single int log level..."test-irv2"

@junhaoliao, if we do need string log level, is there a test file for it?

## Summary by CodeRabbit

- **New Features**
	- Introduced new log event types and enhanced filtering capabilities.
	- Added a new `StructuredIrUnitHandler` for improved log event processing.
	- Updated `ReaderOptions` structure to include `logLevelKey`.

- **Bug Fixes**
	- Enhanced error handling for version validation in stream creation.

- **Refactor**
	- Simplified `filter_log_events` and `decode_range` methods for better maintainability.
	- Removed the `IrUnitHandler` class to streamline log event handling.
	- Cleaned up unused code and header files.

- **Chores**
	- Updated project version to 0.4.0 and added new source files in the build configuration.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1208768426477989
  - https://app.asana.com/0/0/1206305234958554

### DIFF
--- a/src/clp_ffi_js/ir/StreamReader.hpp
+++ b/src/clp_ffi_js/ir/StreamReader.hpp
@@ -101,7 +101,7 @@ public:
      * Deserializes all log events in the stream.
      *
      * @return The number of successfully deserialized ("valid") log events.
-     * @throw ClpFfiJsException if an error occurs deserialization.
+     * @throw ClpFfiJsException if an error occurs during deserialization.
      */
     [[nodiscard]] virtual auto deserialize_stream() -> size_t = 0;
 


CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ project(
     LANGUAGES
         C
         CXX
-    VERSION 0.3.0
+    VERSION 0.4.0
```
junhaoliao, line 17: Can we do the version bump in another PR?

src/clp_ffi_js/constants.hpp
```diff
@@ -19,6 +19,7 @@ enum class LogLevel : std::uint8_t {
     FATAL,
 };
 constexpr LogLevel cValidLogLevelsBeginIdx{LogLevel::TRACE};
+constexpr LogLevel cValidLogLevelsEndIdx{LogLevel::FATAL};
```
junhaoliao, line 22: Indices bounds are non-inclusive, right?

src/clp_ffi_js/constants.hpp
```diff
@@ -19,6 +19,7 @@ enum class LogLevel : std::uint8_t {
     FATAL,
 };
```
junhaoliao, line 20: Can we add a `Length` as the last item and use it to replace `cValidLogLevelsEndIdx`?
```suggestion
    Length
};
```
That way we don't always have to update `cValidLogLevelsEndIdx` if we ever add more log levels.

src/clp_ffi_js/ir/StructuredIrStreamReader.cpp
```diff
@@ -13,39 +13,42 @@
...
+    auto deserialized_log_events{
+            std::make_shared<std::vector<LogEventWithFilterData<StructuredLogEvent>>>()
+    };
     auto result{StructuredIrDeserializer::create(
             *zstd_decompressor,
-            IrUnitHandler{
+            StructuredIrUnitHandler{
                     deserialized_log_events,
+                    reader_options[cReaderOptionsLogLevelKey.data()].as<std::string>(),
                     reader_options[cReaderOptionsTimestampKey.data()].as<std::string>()
```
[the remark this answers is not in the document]
junhaoliao: Let's create a issue for JavaScript support (usage without the TS definition file) and keep a list of TODOs there.

src/clp_ffi_js/ir/utils.hpp
```diff
@@ -0,0 +1,53 @@
...
+ * @param log_level_filter Array of selected log levels.
+ * @param filtered_log_event_map A reference to `FilteredLogEventsMap` that stores filtered result.
+ * @param log_events
+ */
+template <typename LogEvents>
+auto filter_deserialized_events(
+        FilteredLogEventsMap& filtered_log_event_map,
+        LogLevelFilterTsType const& log_level_filter,
+        LogEvents const& log_events
+) -> void {
```
junhaoliao, line 29: Since the results will be stored in `filtered_log_event_map`, shall we move it to the end of the argument list and state that's an output parameter in the docstring?

src/clp_ffi_js/ir/StructuredIrUnitHandler.hpp
junhaoliao, line 1: I hope I didn't overlook, in this file are there any other changes besides these?
1. Adding member `std::string m_log_level_key;` and handling its initialization in the constructor.
2. Adding member `schema_tree_node_id_t m_log_level_node_id;`.
3. Storing `LogEventWithFilterData<StructuredLogEvent>` in the vector instead of `StructuredLogEvent`.

src/clp_ffi_js/ir/StructuredIrUnitHandler.hpp
```diff
@@ -0,0 +1,118 @@
...
+     * @return IRErrorCode::IRErrorCode_Success
+     */
+    [[nodiscard]] static auto handle_utc_offset_change(
+            [[maybe_unused]] clp::UtcOffset utc_offset_old,
+            [[maybe_unused]] clp::UtcOffset utc_offset_new
+    ) -> clp::ffi::ir_stream::IRErrorCode {
+        SPDLOG_WARN("UTC offset change packets aren't handled currently.");
+
+        return clp::ffi::ir_stream::IRErrorCode::IRErrorCode_Success;
+    }
```
junhaoliao, line 66: @kirkrodrigues Do we have any guidelines about separating the method body from its declaration when we have both `.cpp` + `.hpp` files for the class? If so, shall we move this and also the constructor body to the cpp file?

src/clp_ffi_js/ir/StructuredIrUnitHandler.cpp
```diff
@@ -0,0 +1,145 @@
...
+    LogLevel log_level{LogLevel::NONE};
+
+    // Convert the string to uppercase.
+    std::string log_level_name_upper_case{str};
+    std::ranges::transform(
+            log_level_name_upper_case.begin(),
+            log_level_name_upper_case.end(),
+            log_level_name_upper_case.begin(),
+            [](unsigned char c) { return std::toupper(c); }
+    );
```
junhaoliao, line 43: It is good to handle where the log levels are stored fully or partially in lower cases but this is uncommon. Shall we consider removing this for now and add the handling when we implement the new format string syntax?

src/clp_ffi_js/ir/StructuredIrUnitHandler.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+    auto const* it = std::ranges::find(
+            cLogLevelNames.begin() + clp::enum_to_underlying_type(cValidLogLevelsBeginIdx),
+            cLogLevelNames.end(),
+            log_level_name_upper_case
+    );
+
+    if (it == cLogLevelNames.end()) {
+        return log_level;
+    }
```
junhaoliao, line 53: Would this be easier to read?
```suggestion
    if (it == cLogLevelNames.end()) {
        return LogLevel::NONE;
    }
```
And below we can either return the casted log level inline or wrap it with a declaration then return.

src/clp_ffi_js/ir/StructuredIrUnitHandler.cpp
```diff
@@ -0,0 +1,145 @@
...
+#include <clp_ffi_js/ir/LogEventWithFilterData.hpp>
+
+namespace clp_ffi_js::ir {
+
+namespace {
+/**
+ * Parses the `LogLevel` from an input string.
+ * @param str
+ * @return
+ */
```
junhaoliao, line 30:
```suggestion
/**
 * Parses a string to determine the corresponding `LogLevel` enumeration value.
 * 
 * @param str It should match one of the valid log level names in `cLogLevelNames`.
 * @return The parsed value if the input is valid.
 *         If invalid, returns `LogLevel::NONE`.
 */
```

src/clp_ffi_js/ir/StructuredIrUnitHandler.cpp
```diff
@@ -0,0 +1,145 @@
...
+auto parse_log_level(std::string_view str) -> LogLevel;
+
+auto parse_log_level(std::string_view str) -> LogLevel {
```
junhaoliao, line 33: We'll hear from @kirkrodrigues whether we want need forward declarations for helpers like this.

src/clp_ffi_js/ir/StructuredIrUnitHandler.cpp
```diff
@@ -0,0 +1,145 @@
...
+        return log_level;
+    }
+
+    auto const& log_level_pair{id_value_pairs.at(m_log_level_node_id.value())};
+
+    if (false == log_level_pair.has_value()) {
+        return log_level;
+    }
+
+    if (log_level_pair->is<std::string>()) {
```
junhaoliao, line 101: Would this look cleaner?
```suggestion
    if (false == m_log_level_node_id.has_value()) {
        return LogLevel::NONE;
    }

    auto const& log_level_pair{id_value_pairs.at(m_log_level_node_id.value())};
    if (false == log_level_pair.has_value()) {
        return LogLevel::NONE;
    }
    
    LogLevel log_level{LogLevel::NONE};
    if (log_level_pair->is<std::string>()) {
```

src/clp_ffi_js/ir/StructuredIrUnitHandler.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+    if (false == m_log_level_node_id.has_value()) {
+        return log_level;
+    }
+
+    auto const& log_level_pair{id_value_pairs.at(m_log_level_node_id.value())};
+
+    if (false == log_level_pair.has_value()) {
+        return log_level;
+    }
```
junhaoliao, line 99: Would this look cleaner?
```suggestion
    if (false == m_log_level_node_id.has_value()) {
        return LogLevel::NONE;
    }

    auto const& log_level_pair{id_value_pairs.at(m_log_level_node_id.value())};
    if (false == log_level_pair.has_value()) {
        return LogLevel::NONE;
    }
```

src/clp_ffi_js/ir/StructuredIrUnitHandler.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+    if (log_level_pair->is<std::string>()) {
+        auto const& log_level_node_value = log_level_pair.value().get_immutable_view<std::string>();
+        log_level = parse_log_level(log_level_node_value);
```
junhaoliao, line 103:
```suggestion
        return parse_log_level(log_level_node_value);
```

src/clp_ffi_js/ir/StructuredIrUnitHandler.cpp
```diff
@@ -0,0 +1,145 @@
...
+            && log_level_node_value <= clp::enum_to_underlying_type(cValidLogLevelsEndIdx))
+        {
+            log_level = static_cast<LogLevel>(log_level_node_value);
+        }
+    } else {
+        auto log_event_idx = m_deserialized_log_events->size();
+        SPDLOG_ERROR("Log level type is not int or string for log event index {}", log_event_idx);
+    }
+
+    return log_level;
```
junhaoliao, line 117:
```suggestion
    }
    
    SPDLOG_ERROR("Log level type is not int or string for log event index {}", m_deserialized_log_events->size());

    return LogLevel::NONE;
```

junhaoliao (changes requested):
Looks mostly good to me.

src/clp_ffi_js/ir/StructuredIrUnitHandler.cpp
junhaoliao: > Also how is format string re: to this?

Supposedly with the new format string syntax, we can configure how to interpret the raw log event fields.

junhaoliao:
Sorry I still can't make myself available to review this yet. Will hopefully be able to provide some feedback in the next day.

> @junhaoliao, if we do need string log level, is there a test file for it?

I will prepare one and also share the instructions.

src/clp_ffi_js/constants.hpp
junhaoliao: I was looking at https://github.com/y-scope/clp/blob/f1876cd63a56dd79a6243f6c8b6b6a7fd48348cc/components/core/src/clp/GlobalMySQLMetadataDB.cpp#L22-L23 when i suggested the enum name "length" though I don't see the convention being documented anywhere. If that's a convention, i feel we can omit the comment.

@kirkrodrigues could you confirm the convention?

src/clp_ffi_js/ir/StructuredIrStreamReader.cpp
```diff
@@ -119,17 +125,23 @@ auto StructuredIrStreamReader::deserialize_stream() -> size_t {
...
-    m_timestamp_node_id = m_stream_reader_data_context->get_deserializer()
-                                  .get_ir_unit_handler()
-                                  .get_timestamp_node_id();
     m_stream_reader_data_context.reset(nullptr);
     return m_deserialized_log_events->size();
 }
 
 auto StructuredIrStreamReader::decode_range(size_t begin_idx, size_t end_idx, bool use_filter) const
         -> DecodedResultsTsType {
+    if (use_filter && false == m_filtered_log_event_map.has_value()) {
```
junhaoliao, line 134: Some of the logic is duplicated from the UnstructuredIrStreamReader. Do you think it would make sense to extract the common logic into the base class's `decode_range` implementation and let it call a private method of the subclasses? i.e.
1. Let's add a private API `StreamReader::decode_log_event` with a dummy implementation in `StreamReader` that throws "not implemented".
2. Let's override the method in both `StructuredStreamReader` and `UnstructuredStreamReader`.
3. Remove the overrides for `decode_range` in both `StructuredStreamReader` and `UnstructuredStreamReader`.
4. Implement `StreamReader::decode_range` which calls `this->decode_log_event()`.

@kirkrodrigues any objection for this Template Method pattern?

junhaoliao (changes requested):
looks mostly good to me. i proposed some refactoring suggestions and it will be great if @kirkrodrigues can help confirm those and take a further look.
i'll share the instructions to generate an IRv2 file in the morning.

src/clp_ffi_js/constants.hpp
kirkrodrigues: I think it's a general C/C++ convention to call it length; that said, it's probably not a bad idea to leave a comment because maybe the reader would confuse "SIZE" for a log level.

I also realized that the case of these enum values is wrong. They should be `None`, `Trace`, etc. Can we open an issue and fix in another PR?

src/clp_ffi_js/ir/StructuredIrStreamReader.cpp
junhaoliao: I was thinking we could throw a Not Implemented error in the dummy implementation but forgot the cpp compiler should have caught such unimplemented abstract methods in subclasses at compile time. should make it abstract then

src/clp_ffi_js/constants.hpp
junhaoliao: @coderabbit let's open an issue to correct cases of enum property names as PascalCase

src/clp_ffi_js/ir/StructuredIrStreamReader.cpp
junhaoliao: > not straight forward to have a base StreamReader::decode_range() access the derived member variables

true. yup i think moving the the shared code to utils.hpp is a good idea. i imagine a templated helper function can be created if that's what you mean.

src/clp_ffi_js/ir/StructuredIrUnitHandler.cpp
junhaoliao: Right, you got me. i was fuzzily picturing that we could transform the log level during some preprocessing stage of the new formatter, but now with the new syntax implemented i agree that idea doesn't seem to easily / cleanly fit anywhere. Let's keep this

junhaoliao (commented):
overall structure lgtm. I will check the styles and perform a few tests in the evening.

src/clp_ffi_js/ir/StructuredIrStreamReader.cpp
```diff
@@ -94,16 +90,14 @@ auto StructuredIrStreamReader::deserialize_stream() -> size_t {
...
+    while (false == deserializer.is_stream_completed()) {
+        auto result{deserializer.deserialize_next_ir_unit(reader)};
         if (false == result.has_error()) {
             continue;
         }
         auto const error{result.error()};
-        if (std::errc::operation_not_permitted == error) {
-            break;
-        }
         if (std::errc::result_out_of_range == error) {
```
junhaoliao, line 101: @LinZhihao-723 could you help confirm if this is the correct / preferred call sequence to handle End of Stream?

src/clp_ffi_js/ir/StructuredIrUnitHandler.hpp
```diff
@@ -0,0 +1,113 @@
...
+            std::string log_level_key,
+            std::string timestamp_key
+    )
+            : m_log_level_key{std::move(log_level_key)},
+              m_timestamp_key{std::move(timestamp_key)},
+              m_deserialized_log_events{std::move(deserialized_log_events)} {}
+
+    // Methods implementing `clp::ffi::ir_stream::IrUnitHandlerInterface`.
+    /**
+     * Buffers the log event.
```
junhaoliao, line 46: How about
```suggestion
     * Buffers the log event with metadata extracted.
```

src/clp_ffi_js/ir/StructuredIrUnitHandler.cpp
```diff
@@ -0,0 +1,156 @@
...
+        return timestamp;
+    }
+
+    if (timestamp_pair->is<clp::ffi::value_int_t>()) {
+        timestamp = timestamp_pair.value().get_immutable_view<clp::ffi::value_int_t>();
+    } else {
+        // TODO: Add support for parsing timestamp values of string type.
+        auto log_event_idx = m_deserialized_log_events->size();
+        SPDLOG_ERROR("Timestamp type is not int for log event index {}", log_event_idx);
+    }
```
junhaoliao, line 152: more of a preference. feel free to disagree

```suggestion
    auto const& timestamp_pair{id_value_pairs.at(m_timestamp_node_id.value())};
    if (false == timestamp_pair.has_value()) {
        return timestamp;
    }
    if (timestamp_pair->is<clp::ffi::value_int_t>()) {
        timestamp = timestamp_pair.value().get_immutable_view<clp::ffi::value_int_t>();
    } else {
        // TODO: Add support for parsing timestamp values of string type.
        auto log_event_idx = m_deserialized_log_events->size();
        SPDLOG_ERROR("Timestamp type is not int for log event index {}", log_event_idx);
    }
```

src/clp_ffi_js/ir/utils.hpp
```diff
@@ -0,0 +1,165 @@
...
+ * @return
+ */
+template <typename LogEvents>
+auto generic_decode_range(
+        size_t begin_idx,
+        size_t end_idx,
+        FilteredLogEventsMap const& filtered_log_event_map,
+        LogEvents const& log_events,
+        bool use_filter,
+        clp::TimestampPattern const& ts_pattern
```
junhaoliao, line 88: @kirkrodrigues 
1. Shall we use `std::enable_if<std::is_same_v<>>` for `ts_pattern`?
2. if not, do we consider Pack / variadic arguments a bad coding practice in C++? can that be used here?
3. if not, can we create a generic DecoderOptions class?

junhaoliao (changes requested):
all good except one opening discussion for the TimestampPattern parameter in the `generic_decode_range` signature.

src/clp_ffi_js/ir/utils.hpp
```diff
@@ -0,0 +1,163 @@
...
+            }
+        }
+
+        EM_ASM(
+                { Emval.toValue($0).push([UTF8ToString($1), $2, $3, $4]); },
+                results.as_handle(),
+                generic_event_string.c_str(),
+                timestamp,
+                log_level,
+                log_event_idx + 1
```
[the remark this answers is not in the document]
junhaoliao: i'm not sure how your suggestion would help. could you explain a bit?

src/clp_ffi_js/ir/utils.hpp
```diff
@@ -0,0 +1,165 @@
...
+        } else {
+            log_event_idx = i;
+        }
+
+        auto const& log_event_with_filter_data{log_events.at(log_event_idx)};
+        auto const& log_event = log_event_with_filter_data.get_log_event();
+        auto const& timestamp = log_event_with_filter_data.get_timestamp();
+        auto const& log_level = log_event_with_filter_data.get_log_level();
+
+        std::string generic_event_string;
```
junhaoliao, line 119: Can we lift this variable and also the length reservation outside of the for-loop?

src/clp_ffi_js/ir/utils.hpp
kirkrodrigues: 1. Sure
2. clang-tidy [discourages](https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/pro-type-vararg.html) variadic functions.
3. Sure, but how would that help?

src/clp_ffi_js/ir/StructuredIrStreamReader.cpp
LinZhihao-723: - We don't need to check any error code on the end of stream; the end of stream is handled as an IR unit; you can check the returned IR unit type to confirm (but that should also be a part of the job of the unit handler)
- This error code was added to check truncated IR, which is the incomplete stream error. It's been fixed in ffi core, so only `std::errc::result_out_of_range` is needed: https://github.com/y-scope/clp/pull/579

Btw, u can refer to this part of code in clp-ffi-py to see some similar usage: https://github.com/y-scope/clp-ffi-py/blob/main/src/clp_ffi_py/ir/native/PyDeserializer.cpp#L224

src/clp_ffi_js/ir/utils.hpp
```diff
@@ -0,0 +1,163 @@
...
+ * @param end_idx
+ * @param filtered_log_event_map
+ * @param log_events
+ * @param use_filter
+ * @param ts_pattern A pattern for formatting unstructured log event timestamps as date strings.
+ * The pattern is unused for structured log events, as they are not formatted in `clp-ffi-js`.
+ * @return
+ */
+template <typename LogEvents>
+auto generic_decode_range(
```
kirkrodrigues, line 82: I haven't tried this, so maybe there's some unforeseen limitation, but I think a clean approach is:

* Make StreamReader a template class that takes the log event type as a template parameter.
* Add an abstract method to `StreamReader` for converting a log event into a string---this is where the derived classes can implement non-generic part of `generic_decode_range` will go.
* Make `decode_range` and `filter_log_events` non-abstract methods of `StreamReader`, and use the util function implementations as the method implementations.
* In `decode_range`, call the new method you added.

kirkrodrigues (changes requested):

src/clp_ffi_js/ir/utils.hpp
kirkrodrigues: Can we move those into StreamReader? I think the only difference (besides the template parameter) is that one is a shared_ptr and one isn't, right?

src/clp_ffi_js/ir/utils.hpp
junhaoliao: 1. I was thinking we can declare separate function signatures (one that is specialized for class UnstructuredStreamReader and the other one(s) is/are for class StructuredStreamReader and others. For the common logics, both of them call the same helpers.
2. In a way I see how (1.) can be considered a bad practice for similar reasons. It could be hard to maintain when we add more stream reader classes.
3. If I don't picture things wrong, in the StreamReader-type templated DecoderOptions class we can enable attributes only for certain StreamReader types. Something like
    ```cpp
    template<typename StreamReader>
    class DecoderOptions {
    public:
        // Other members can be added here
    };
    
    // Specialization for UnstructuredStreamReader
    template<>
    class DecoderOptions<UnstructuredStreamReader> {
    public:
        clp::TimestampPattern const& ts_pattern;
    
        DecoderOptions(const clp::TimestampPattern& ts_pattern_ref) 
            : ts_pattern(ts_pattern_ref) {}
    };
    ```
    
    i guess we can also do
    ```
    template<typename StreamReader>
    class DecoderOptions {
    public:
        // Constructor specialization for UnstructuredStreamReader
        template<typename T = StreamReader, typename = std::enable_if_t<std::is_same_v<T, UnstructuredStreamReader>>>
        DecoderOptions(const clp::TimestampPattern& ts_pattern_ref) 
            : ts_pattern(&ts_pattern_ref) {}
    
        // or make this std::Optional
        const clp::TimestampPattern* ts_pattern = nullptr;
    };
    ```

junhaoliao: Let's continue the discussion at https://github.com/y-scope/clp-ffi-js/pull/35/files#r1882227286 before we look at the approaches proposed in this thread.

src/clp_ffi_js/ir/utils.hpp
kirkrodrigues: Right, Marco mentioned that we can only embind an instantiated type, so changing decode_range to use a lambda is probably the cleaner approach.

src/clp_ffi_js/constants.hpp
```diff
@@ -17,6 +18,7 @@ enum class LogLevel : std::uint8_t {
     WARN,
     ERROR,
     FATAL,
+    LENGTH,  // This isn't a valid log level.
```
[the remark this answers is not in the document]
kirkrodrigues: I see what you're saying. On the other hand, having two invalid values is kind of messy already. For example, the code to check if a log level is within the valid enum values depends on `cValidLogLevelsBeginIdx` and `LogLevel::LENGTH`.

My perspective on the current log level parsing is that it's a heuristic until we migrate to IRv2 entirelyl; at that point, we shouldn't have to parse log levels any more. So I feel like simplifying the current code is a better approach until we switch to IRv2 or we have a user-need for more complicated levels.

That said, this isn't that big of a deal, so I guess if @junhaoliao doesn't agree, I'll leave it be.

src/clp_ffi_js/ir/StreamReader.hpp
```diff
@@ -100,6 +115,108 @@ class StreamReader {
     [[nodiscard]] virtual auto decode_range(size_t begin_idx, size_t end_idx, bool use_filter) const
             -> DecodedResultsTsType = 0;
 
+    /**
```
kirkrodrigues, line 118: * `generic_filter_log_events` and `generic_decode_range` should be protected, right?
* Can we move the implementations after the class declaration?

src/clp_ffi_js/ir/StreamReader.hpp
```diff
@@ -100,6 +115,108 @@ class StreamReader {
...
+     * A templated function that implements `filter_log_events` for both
+     * `UnstructuredIrStreamReader` and `StructuredIrStreamReader`.
```
kirkrodrigues, line 120:
```suggestion
     * Templated implementation of `filter_log_events`.
```

src/clp_ffi_js/ir/StreamReader.hpp
```diff
@@ -100,6 +115,108 @@ class StreamReader {
...
+     *
+     * @param log_level_filter
+     * @param log_events Derived class's log events.
+     * @param[out] filtered_log_event_map A reference to derived class's `FilteredLogEventsMap`
+     * that stores filtered result.
```
kirkrodrigues, line 125:
```suggestion
     * @param log_events
     * @param[out] filtered_log_event_map Returns the filtered log events.
```

We use "Returns" for the parameter according to our [internal guidelines](https://www.notion.so/yscope/WIP-Coding-Guidelines-9a308b847a5343958ba3cb97a850be66?pvs=4#9c18b7d2cf2543dea6a3b56f05072287).

src/clp_ffi_js/ir/StreamReader.hpp
```diff
@@ -100,6 +115,108 @@ class StreamReader {
...
+     */
+    template <typename LogEvents>
+    static auto generic_filter_log_events(
+            FilteredLogEventsMap& filtered_log_event_map,
+            LogLevelFilterTsType const& log_level_filter,
+            LogEvents const& log_events
```
kirkrodrigues, line 131:
```suggestion
    template <typename LogEvent>
    static auto generic_filter_log_events(
            FilteredLogEventsMap& filtered_log_event_map,
            LogLevelFilterTsType const& log_level_filter,
            std::vector<LogEventWithFilterData<LogEvent>> const& log_events
```

This is more explicit with the requirements for the type.

src/clp_ffi_js/ir/StreamReader.hpp
```diff
@@ -100,6 +115,108 @@ class StreamReader {
...
+                != filter_levels.end())
+            {
+                filtered_log_event_map->emplace_back(log_event_idx);
+            }
+        }
+    }
+
+    /**
+     * A templated function that implements `decode_range` for both
+     * `UnstructuredIrStreamReader` and `StructuredIrStreamReader`.
```
kirkrodrigues, line 157:
```suggestion
     * Templated implementation of `decode_range` that uses `log_event_to_string` to convert
     * `log_event` to a string for the returned result.
```

src/clp_ffi_js/ir/StreamReader.hpp
```diff
@@ -100,6 +115,108 @@ class StreamReader {
     [[nodiscard]] virtual auto decode_range(size_t begin_idx, size_t end_idx, bool use_filter) const
             -> DecodedResultsTsType = 0;
 
+    /**
+     * A templated function that implements `filter_log_events` for both
+     * `UnstructuredIrStreamReader` and `StructuredIrStreamReader`.
+     *
+     * @param log_level_filter
```
kirkrodrigues, line 122:
```suggestion
     * @tparam LogEvent
     * @param log_level_filter
```

src/clp_ffi_js/ir/StreamReader.hpp
```diff
@@ -100,6 +115,108 @@ class StreamReader {
...
+                filtered_log_event_map->emplace_back(log_event_idx);
+            }
+        }
+    }
+
+    /**
+     * A templated function that implements `decode_range` for both
+     * `UnstructuredIrStreamReader` and `StructuredIrStreamReader`.
+     *
+     * @param begin_idx
```
kirkrodrigues, line 159:
```suggestion
     * @tparam LogEvent
     * @tparam ToStringFunc Function to convert a log event into a string.
     * @param begin_idx
```

src/clp_ffi_js/ir/StreamReader.hpp
```diff
@@ -100,6 +115,108 @@ class StreamReader {
...
+     * @param end_idx
+     * @param filtered_log_event_map Derived class's `FilteredLogEventsMap`.
+     * @param log_events Derived class's log events.
+     * @param use_filter
+     * @param log_event_to_string Lambda function which retrieves a string from a single log event.
+     * The derived class must implement lambda since `Unstructured` and `Structured`
+     * log events have different interfaces.
+     * @return
```
kirkrodrigues, line 167:
```suggestion
     * @param filtered_log_event_map
     * @param log_events
     * @param use_filter
     * @param log_event_to_string
     * @return See `decode_range`.
```

src/clp_ffi_js/ir/StreamReader.hpp
```diff
@@ -100,6 +115,108 @@ class StreamReader {
...
+     */
+    template <typename LogEvents, typename ToStringFunc>
+    static auto generic_decode_range(
+            size_t begin_idx,
+            size_t end_idx,
+            FilteredLogEventsMap const& filtered_log_event_map,
+            LogEvents const& log_events,
```
kirkrodrigues, line 174:
```suggestion
    template <typename LogEvent, typename ToStringFunc>
    requires requires(ToStringFunc func, LogEvent const& log_event) {
        {
            func(log_event)
        } -> std::convertible_to<std::string>;
    }
    static auto generic_decode_range(
            size_t begin_idx,
            size_t end_idx,
            FilteredLogEventsMap const& filtered_log_event_map,
            std::vector<LogEventWithFilterData<LogEvent>> const& log_events,
```

This constrains the types more explicitly.

src/clp_ffi_js/ir/StructuredIrStreamReader.cpp
```diff
@@ -75,15 +68,19 @@ auto StructuredIrStreamReader::get_num_events_buffered() const -> size_t {
...
+
+    return FilteredLogEventMapTsType{emscripten::val::array(m_filtered_log_event_map.value())};
 }
 
 void StructuredIrStreamReader::filter_log_events(LogLevelFilterTsType const& log_level_filter) {
-    if (log_level_filter.isNull()) {
-        return;
-    }
-    SPDLOG_ERROR(cLogLevelFilteringNotSupportedErrorMsg);
+    StreamReader::generic_filter_log_events(
```
kirkrodrigues, line 79:
```suggestion
    generic_filter_log_events(
```

src/clp_ffi_js/ir/StructuredIrStreamReader.cpp
```diff
@@ -119,72 +114,42 @@ auto StructuredIrStreamReader::deserialize_stream() -> size_t {
...
-                results.as_handle(),
-                json_str.c_str(),
-                timestamp,
-                LogLevel::NONE,
-                log_event_idx + 1
-        );
-    }
-
-    return DecodedResultsTsType(results);
+    return StreamReader::generic_decode_range(
```
kirkrodrigues, line 140:
```suggestion
    return generic_decode_range(
```

src/clp_ffi_js/ir/StructuredIrUnitHandler.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+auto StructuredIrUnitHandler::get_log_level(
+        StructuredLogEvent::NodeIdValuePairs const& id_value_pairs
+) const -> LogLevel {
+    LogLevel log_level{LogLevel::NONE};
+
+    if (false == m_log_level_node_id.has_value()) {
+        return log_level;
+    }
+    auto const& log_level_pair{id_value_pairs.at(m_log_level_node_id.value())};
```
kirkrodrigues, line 104: I thought I posted this in the previous review but it's not there so I guess I didn't post it or GH lost it before posting. Anyway-

This isn't a pair, right? This is the value from the `id_value_pairs` entry. Since the value itself is an optional, how about `optional_log_level_value`? Then after checking that the optional isn't empty, we can add a variable `log_level_value` and use that in the rest of the method.

We can do the same thing in `get_timestamp`.

src/clp_ffi_js/ir/UnstructuredIrStreamReader.cpp
```diff
@@ -74,25 +69,11 @@ auto UnstructuredIrStreamReader::get_filtered_log_event_map() const -> FilteredL
...
-        if (std::ranges::find(
-                    filter_levels,
-                    clp::enum_to_underlying_type(log_event.get_log_level())
-            )
-            != filter_levels.end())
-        {
-            m_filtered_log_event_map->emplace_back(log_event_idx);
-        }
-    }
+    StreamReader::generic_filter_log_events(
```
kirkrodrigues, line 72:
```suggestion
    generic_filter_log_events(
```

src/clp_ffi_js/ir/UnstructuredIrStreamReader.cpp
```diff
@@ -196,20 +148,18 @@ auto UnstructuredIrStreamReader::decode_range(size_t begin_idx, size_t end_idx,
...
-                log_event_idx + 1
-        );
-    }
-
-    return DecodedResultsTsType(results);
+        m_ts_pattern.insert_formatted_timestamp(log_event.get_timestamp(), message);
+        return message;
+    };
+
+    return StreamReader::generic_decode_range(
```
kirkrodrigues, line 155:
```suggestion
    return generic_decode_range(
```

src/clp_ffi_js/ir/StructuredIrUnitHandler.cpp
```diff
@@ -0,0 +1,161 @@
...
+        return log_level;
+    }
+
+    auto const log_level_value = log_level_optional_value.value();
+
+    if (log_level_value.is<std::string>()) {
+        auto const& log_level_str = log_level_value.get_immutable_view<std::string>();
+        log_level = parse_log_level(log_level_str);
+    } else if (log_level_value.is<clp::ffi::value_int_t>()) {
+        auto const& log_level_int = (log_level_value.get_immutable_view<clp::ffi::value_int_t>());
```
kirkrodrigues, line 115:
```suggestion
        auto const& log_level_int = log_level_value.get_immutable_view<clp::ffi::value_int_t>();
```

src/clp_ffi_js/ir/StructuredIrUnitHandler.cpp
```diff
@@ -0,0 +1,161 @@
...
+    LogLevel log_level{LogLevel::NONE};
+
+    if (false == m_log_level_node_id.has_value()) {
+        return log_level;
+    }
+    auto const& log_level_optional_value{id_value_pairs.at(m_log_level_node_id.value())};
+    if (false == log_level_optional_value.has_value()) {
+        return log_level;
+    }
+
```
kirkrodrigues, line 108: Nit:
```suggestion
```

src/clp_ffi_js/ir/StructuredIrUnitHandler.cpp
```diff
@@ -0,0 +1,161 @@
...
+auto StructuredIrUnitHandler::get_log_level(
+        StructuredLogEvent::NodeIdValuePairs const& id_value_pairs
+) const -> LogLevel {
+    LogLevel log_level{LogLevel::NONE};
+
+    if (false == m_log_level_node_id.has_value()) {
+        return log_level;
+    }
+    auto const& log_level_optional_value{id_value_pairs.at(m_log_level_node_id.value())};
```
kirkrodrigues, line 104: Although what you wrote is grammatically correct, we have an [internal guideline](https://www.notion.so/yscope/WIP-Coding-Guidelines-9a308b847a5343958ba3cb97a850be66?pvs=4#eb038f1b23cf4f2098859025061a1d05) that requires the name is `optional_xxx`. Unless you're purposefully violating that guideline?

kirkrodrigues: Same comment for `timestamp_optional_value`.

kirkrodrigues (changes requested):
A few more minor comments.

kirkrodrigues (dismissed):
For the PR title, how about:

```
feat: Add support for log-level filtering of structured IR streams.
```

src/clp_ffi_js/constants.hpp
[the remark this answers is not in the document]
junhaoliao: > the current log level parsing is that it's a heuristic until we migrate to IRv2 entirely

@kirkrodrigues Would CLP FFI also support parsing log level for IRv1 stream in the new `Deserializer` interface (the one that is now only for IRv2)? If not, we still need to maintain this enum for IRv1 till we completely drop support for IRv1, right?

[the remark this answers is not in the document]
kirkrodrigues: Yeah, we need to wait until we completely drop support for IRv1.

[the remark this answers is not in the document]
junhaoliao: I agree with @davemarco 's (1.) argument about the challenges of adding new log levels. However, I also recognize that the `NONE` value is used solely for log parsing (in read paths, not write paths). Since `NONE` is never serialized into non-code spaces and read back into the code, as long as our implementation consistently refers to it, there should be no risk of errors, which aligns with @kirkrodrigues's argument, assuming I understand it correctly.

With that said, I still prefer that we keep the `NONE` as the very first item in the enum.
The underlying type of the `LogLevel` is `uint8_t` which has value `0` when unutilized and placed in bss / initialized without explicit value (`={}`). Such value is usually used to hint the default; for example, in order to hint feature XXX is by default enabled, we can add Boolean flags like `is_xxx_disabled`. With `NONE` removed, when we declare a enum variable without proper initialization, it may contain a valid value, and most importantly the default value is no longer hinted to code readers. (Though most of the time we shouldn't really read before initialization lol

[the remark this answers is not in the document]
junhaoliao: In a way I feel the inline comments seem a bit bloated. Instead, as a new convention, can we add a prefix to hint the enum member is for describing properties of the enum? e.g., 
```
enum class FilesTableCol: std::uint8_t {
   NAME = 0,
   PATH,
   SIZE,
   FILES_TABLE_COL_LENGTH // then this should not be confused with `SIZE`
}
```

src/clp_ffi_js/ir/StreamReader.hpp
```diff
@@ -102,7 +123,136 @@ class StreamReader {
...
+auto StreamReader::generic_decode_range(
+        size_t begin_idx,
+        size_t end_idx,
+        FilteredLogEventsMap const& filtered_log_event_map,
+        LogEvents<LogEvent> const& log_events,
+        ToStringFunc log_event_to_string,
+        bool use_filter
+) -> DecodedResultsTsType {
+    if (use_filter && false == filtered_log_event_map.has_value()) {
+        return DecodedResultsTsType{emscripten::val::null()};
```
junhaoliao, line 188: this isn't in the log viewer code as well -

Can we add a log print before returning so that we know the input argument check fails here rather than at the range check?

src/clp_ffi_js/ir/StreamReader.hpp
```diff
@@ -102,7 +123,136 @@ class StreamReader {
...
+    }
+
+    size_t length{0};
+    if (use_filter) {
+        length = filtered_log_event_map->size();
+    } else {
+        length = log_events.size();
+    }
+    if (length < end_idx || begin_idx > end_idx) {
+        return DecodedResultsTsType{emscripten::val::null()};
```
junhaoliao, line 198: see https://github.com/y-scope/clp-ffi-js/pull/35/files#r1884668582

it be a good idea to add a log print here before returning

junhaoliao (approved):
test with a clp-ffi-py generated IRv2 file with various log levels. all good except
1. logs can be printed before returning null in `auto StreamReader::generic_decode_range`: https://github.com/y-scope/clp-ffi-js/pull/35/files#r1884668582
2. a pending discussion at https://github.com/y-scope/clp-ffi-js/pull/35/files#r1881142029

src/clp_ffi_js/constants.hpp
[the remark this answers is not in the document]
kirkrodrigues: * Sure, let's keep `NONE`.
* Sure, I guess we could define a new convention. Keep in mind the naming of enum values should be PascalCase.

src/clp_ffi_js/ir/StructuredIrUnitHandler.cpp
```diff
@@ -0,0 +1,160 @@
...
+        SPDLOG_ERROR(
+                "Authoritative log level's value is not an int or string for log event index {}",
+                log_event_idx
+        );
+    }
+
+    return log_level;
+}
+
+auto StructuredIrUnitHandler::get_timestamp(
```
junhaoliao, line 131: This may be deduplicated with `StructuredIrUnitHandler::get_log_level`, but for the sake of simplicity let's leave this for now.

src/clp_ffi_js/ir/UnstructuredIrStreamReader.cpp
```diff
@@ -155,38 +132,9 @@ auto UnstructuredIrStreamReader::deserialize_stream() -> size_t {
...
-        auto const& unstructured_log_event = log_event_with_filter_data.get_log_event();
-        auto const& log_level = log_event_with_filter_data.get_log_level();
-        auto const& timestamp = log_event_with_filter_data.get_timestamp();
-
-        auto const parsed{unstructured_log_event.get_message().decode_and_unparse()};
+    auto log_event_to_string = [this](UnstructuredLogEvent const& log_event) -> std::string {
+        std::string message;
+        auto const parsed{log_event.get_message().decode_and_unparse()};
         if (false == parsed.has_value()) {
             throw ClpFfiJsException{
```
junhaoliao, line 139: Since the structured one also throws the same error when a message cannot be decoded, let's document this `@throw` in the docstring at https://github.com/y-scope/clp-ffi-js/blob/d2556b2fb6d986182b247c660244dfbc923e37ea/src/clp_ffi_js/ir/StreamReader.hpp#L100-L101

junhaoliao (approved):
Added some comments about documentation for the exception behaviour

For the PR title, how about
```
feat(StructuredIrStreamReader): Add support for log-level filtering.
```

src/clp_ffi_js/ir/StreamReader.hpp
```diff
@@ -102,7 +124,137 @@ class StreamReader {
...
+     *
+     * @tparam LogEvent
+     * @tparam ToStringFunc Function to convert a log event into a string.
+     * @param begin_idx
+     * @param end_idx
+     * @param filtered_log_event_map
+     * @param log_events
+     * @param use_filter
+     * @param log_event_to_string
+     * @return See `decode_range`.
```
junhaoliao, line 140: Sorry for the lack of context at https://github.com/y-scope/clp-ffi-js/pull/35#discussion_r1888811976 

```suggestion
     * @return See `decode_range`.
     * @throw Propagates `ToStringFunc`'s exception.
```

src/clp_ffi_js/ir/StreamReader.hpp
```diff
@@ -102,7 +124,138 @@ class StreamReader {
...
+     * @throw Propagates exceptions from `ToStringFunc`.
```
junhaoliao, line 141: The suggested wording was from the Coding Guidelines. Might be good to keep that for consistency: https://www.notion.so/yscope/WIP-Coding-Guidelines-9a308b847a5343958ba3cb97a850be66?pvs=4#13504e4d9e6b80d3bc29c87e930fcee1

src/clp_ffi_js/ir/UnstructuredIrStreamReader.cpp
junhaoliao: Sorry for the confusion. I mean we want to add this at https://github.com/y-scope/clp-ffi-js/blob/d2556b2fb6d986182b247c660244dfbc923e37ea/src/clp_ffi_js/ir/StreamReader.hpp#L99
```
@throw ClpFfiJsException if any error occurs.
```
or this may be more clear:
```
@throw ClpFfiJsException if a message cannot be decoded.
```

junhaoliao:
> re: title I think the current title is better. I think adding `StructuredIrStreamReader` to the title is too much detail for someone not familiar with code. Maybe `feat(structured): Add support for log-level filtering of structured IR streams`? Or just leave as is.

Right, the scope is optional in Conventional Commit messages. Sure let's leave the PR title as is

junhaoliao (approved):

PR #42, opened 2024-12-19
feat: Add support for finding the log event that's closest to a target timestamp.

Henry8192:
# Description
Adds `getLogEventIndexByTimestamp` API in `clp_ffi_js/ir/StreamReader.cpp` for ClpStreamReader. It returns `null` only if there are no log events. Otherwise, it would search logs in "best effort", which means it returns the last index smaller than the timestamp if no exact timestamp match, unless all log event timestamps are larger than the target. In that case, return the first log event index.

In `clp_ffi_js/ir/StreamReader.hpp`, implements `generic_get_log_event_index_by_timestamp()`, the core logic (basically just binary search) for finding log events by timestamp.
This function is called both by `StructuredIrStreamReader` and `UnstructuredIrStreamReader`'s `get_log_event_index_by_timestamp`.

# Validation performed
Use the following javascript for testing structured and unstructured logs:
```js
import ModuleInit from "./cmake-build-debug/ClpFfiJs-node.js"
import fs from "node:fs"

const main = async () => {
    const file = fs.readFileSync("./test-1K.clp.zst")

    console.time("perf")
    const Module = await ModuleInit()
    try {
        const decoder = new Module.ClpStreamReader(new Uint8Array(file), {logLevelKey: "$log_level", timestampKey: "$timestamp"})
        console.log("type:", decoder.getIrStreamType() === Module.IrStreamType.STRUCTURED ? "structured" : "unstructured")
        const numEvents = decoder.deserializeStream()
        console.log(numEvents)
        const results = decoder.decodeRange(0, numEvents, false)
        console.log(results)
        const logEventIdx = decoder.getLogEventIndexByTimestamp(1736183181284)
        console.log(logEventIdx)
    } catch (e) {
        console.error("Exception caught:", e.stack)
    }
    console.timeEnd("perf")
}

void main()
```

irv2 logs generating scripts:
```py
"""
To install all dependencies:
    pip install clp-ffi-py==0.1.0b1 msgpack nabg zstandard
TODO: set random seed based on `i` so that all logs generated are deterministic.
"""
import random
import time

import msgpack
import nabg
import zstandard as zstd
from clp_ffi_py.ir.native import Serializer

NUM_EVENTS_TO_GENERATE = 1_000
FILE_NAME = "test-1K.clp.zst"

LOG_LEVELS = ["TRACE", 'DEBUG', 'INFO', 'WARN', 'ERROR', 'FATAL', "SOME_UNKNOWN_LEVEL"]

EMOJI_RANGES = [
    (0x1F600, 0x1F64F),  # Emoticons
    (0x1F300, 0x1F5FF),  # Miscellaneous Symbols and Pictographs
    (0x1F680, 0x1F6FF),  # Transport and Map Symbols
    (0x1F900, 0x1F9FF),  # Supplemental Symbols and Pictographs
    (0x2600, 0x26FF),  # Miscellaneous Symbols
    (0x1F1E6, 0x1F1FF),  # Flags
]
EMOJI_POOL = [chr(i) for r in EMOJI_RANGES for i in range(r[0], r[1] + 1)]

NUM_MILLIS_IN_SECOND = 1000


def generate_random_emoji():
    return random.choice(EMOJI_POOL)


def generate_random_message():
    return (
            generate_random_emoji() +
            nabg.ionize() +
            generate_random_emoji()
    )


cctx = zstd.ZstdCompressor(threads=2)

with open(FILE_NAME, "wb") as f:
    with cctx.stream_writer(f) as w:
        with Serializer(w) as s:
            for i in range(NUM_EVENTS_TO_GENERATE):
                random.seed(i)
                log_event = msgpack.dumps({
                    "$log_level": LOG_LEVELS[i % len(LOG_LEVELS)],
                    "$timestamp": int(time.time() * NUM_MILLIS_IN_SECOND),
                    "message": generate_random_message()
                })
                s.serialize_log_event_from_msgpack_map(log_event)
```

## Summary by CodeRabbit

- **New Features**
	- Introduced a capability to locate the nearest log event index based on a specified timestamp across key streaming components.
	- Enhanced log event indexing with a new flexible data type that supports both numerical values and the absence of a value, ensuring more robust and efficient data handling.


junhaoliao (changes requested):
For future commits, you may follow https://github.com/y-scope/clp-ffi-js?tab=readme-ov-file#linting to run the linter

src/clp_ffi_js/ir/StructuredIrStreamReader.cpp
```diff
@@ -147,6 +147,30 @@ auto StructuredIrStreamReader::decode_range(size_t begin_idx, size_t end_idx, bo
...
+            m_deserialized_log_events->end(),
+            input_timestamp,
+            [](const LogEventWithFilterData<StructuredLogEvent>& event, clp::ir::epoch_time_ms_t timestamp) {
+                return event.get_timestamp() <= timestamp;
+            }
+    );
+
+    // Adjust the iterator to find the last valid index
+    if (it == m_deserialized_log_events->end() || it->get_timestamp() > input_timestamp) {
+        if (it == m_deserialized_log_events->begin()) {
```
junhaoliao, line 165: Is it possible to check against `m_deserialized_log_events->empty()` at the beginning of the method and early return, so that we can avoid checking `it == m_deserialized_log_events->end() && it == m_deserialized_log_events->begin()`?

src/clp_ffi_js/ir/StructuredIrStreamReader.cpp
```diff
@@ -147,6 +147,30 @@ auto StructuredIrStreamReader::decode_range(size_t begin_idx, size_t end_idx, bo
...
+            return -1;  // No element satisfies the condition
+        }
+        --it;
```
junhaoliao, line 168: Can we use `std::range::upper_bound` instead and always `it--` after, so that we can avoid checking `(it == m_deserialized_log_events->end() || it->get_timestamp() > input_timestamp)`?

src/clp_ffi_js/ir/UnstructuredIrStreamReader.cpp
```diff
@@ -158,6 +158,29 @@ auto UnstructuredIrStreamReader::decode_range(size_t begin_idx, size_t end_idx,
     );
 }
 
+auto UnstructuredIrStreamReader::find_timestamp_last_occurrence(
```
junhaoliao, line 161: Let's deduplicate this with `StructuredIrStreamReader::find_timestamp_last_occurrence`.
You may refer to `generic_decode_range`.

junhaoliao (changes requested):

src/clp_ffi_js/ir/StreamReader.hpp
junhaoliao, line 1: For future commits, you may follow https://github.com/y-scope/clp-ffi-js?tab=readme-ov-file#linting to run the linter

src/clp_ffi_js/ir/UnstructuredIrStreamReader.hpp
junhaoliao, line 1: Let's `#include <clp/ir/types.hpp>` so that the header providing the type can be directly included.

src/clp_ffi_js/ir/StructuredIrStreamReader.hpp
junhaoliao, line 1: Let's `#include <clp/ir/types.hpp>` so that the header providing the type can be directly included.

src/clp_ffi_js/ir/UnstructuredIrStreamReader.hpp
Henry8192: I think it is already included on line 9

junhaoliao:
This PR is blocking https://github.com/y-scope/yscope-log-viewer/pull/152

src/clp_ffi_js/ir/StreamReader.hpp
```diff
@@ -258,6 +289,42 @@ auto StreamReader::generic_filter_log_events(
...
+                 LogEventWithFilterData<LogEvent> const& event,
+                 clp::ir::epoch_time_ms_t timestamp
+         ) {
+    {
+        event.get_timestamp()
+    } -> std::convertible_to<clp::ir::epoch_time_ms_t>;
+}
+auto StreamReader::generic_get_log_event_index_by_timestamp(
+        Iterator begin,
+        Iterator end,
```
junhaoliao, line 304: Can we pass in the vector directly? e.g. `LogEvents<LogEvent> const& log_events`

junhaoliao (changes requested):

src/clp_ffi_js/ir/StreamReader.cpp
```diff
@@ -129,6 +129,7 @@ EMSCRIPTEN_BINDINGS(ClpStreamReader) {
             "Array<[string, bigint, number, number]>"
     );
     emscripten::register_type<clp_ffi_js::ir::FilteredLogEventMapTsType>("number[] | null");
+    emscripten::register_type<clp_ffi_js::ir::LogEventIdxTsType>("number | null");
```
junhaoliao, line 132: Registering a type that is nullable is good for consistency with existing code like `FilteredLogEventMapTsType`.

That said, Embind seems to have added support for `std::optional<>` so we should look into converting functions that return nullable data into returning `std::optional<>` types.

https://emscripten.org/docs/porting/connecting_cpp_and_javascript/embind.html#:~:text=For%20convenience%2C%20embind%20provides%20factory%20functions%20to%20register%20std%3A%3Avector%3CT%3E%20(register_vector())%2C%20std%3A%3Amap%3CK%2C%20V%3E%20(register_map())%2C%20and%20std%3A%3Aoptional%3CT%3E%20(register_optional())%20types%3A
> For convenience, embind provides factory functions to register std::vector<T> ([register_vector()](https://emscripten.org/docs/api_reference/bind.h.html#_CPPv415register_vectorPKc)), std::map<K, V> ([register_map()](https://emscripten.org/docs/api_reference/bind.h.html#_CPPv412register_mapPKc)), and std::optional<T> (register_optional()) types:

@coderabbitai create an issue to track this

src/clp_ffi_js/ir/StreamReader.hpp
```diff
@@ -258,6 +295,39 @@ auto StreamReader::generic_filter_log_events(
...
+                return ts < log_event.get_timestamp();
+            }
+    );
+
+    // If the iterator is not pointing to the beginning, decrement it so that it points to
+    // first element that is larger than timestamp.
+    if (it != log_events.begin()) {
+        --it;
+    }
+    return LogEventIdxTsType{emscripten::val(std::distance(log_events.begin(), it))};
```
junhaoliao, line 329: how about we rename the `it` as `upper` then this code becomes self-explanatory.

```suggestion
    if (upper == log_events.begin()) {
        return LogEventIdxTsType{0};
    }

    size_t upper_index{std::distance(log_events.begin(), upper)};
    const auto index{upper_index - 1};

    return LogEventIdxTsType{emscripten::val(index)};
```

junhaoliao (changes requested):

src/clp_ffi_js/ir/StreamReader.hpp
```diff
@@ -163,7 +177,6 @@ class StreamReader {
      * @tparam LogEvent
      * @param log_level_filter
      * @param log_events Derived class's log events.
-     * @param log_events
```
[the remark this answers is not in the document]
Henry8192: isn't it duplicated?

src/clp_ffi_js/ir/StreamReader.hpp
```diff
@@ -258,6 +295,39 @@ auto StreamReader::generic_filter_log_events(
...
+        return LogEventIdxTsType{emscripten::val::null()};
+    }
+    auto it = std::upper_bound(
+            log_events.begin(),
+            log_events.end(),
+            timestamp,
+            [](clp::ir::epoch_time_ms_t ts, LogEventWithFilterData<LogEvent> const& log_event) {
+                return ts < log_event.get_timestamp();
+            }
+    );
```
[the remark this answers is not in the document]
Henry8192: I don't think this syntax would work.

src/clp_ffi_js/ir/StreamReader.hpp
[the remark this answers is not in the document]
junhaoliao: Let's revert the change and fix it in another PR. I think we should also add docs for `filtered_log_event_map`.

src/clp_ffi_js/ir/StreamReader.cpp
junhaoliao: > we should look into converting functions...

in a separate PR i meant, so we should create an issue to track the effort

src/clp_ffi_js/ir/StreamReader.hpp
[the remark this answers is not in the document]
junhaoliao: it's a space off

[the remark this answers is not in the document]
Henry8192: lol you are right

src/clp_ffi_js/ir/StreamReader.hpp
```diff
@@ -44,6 +46,23 @@ using LogEvents = std::vector<LogEventWithFilterData<LogEvent>>;
...
+    } -> std::convertible_to<clp::ir::epoch_time_ms_t>;
+};
+
+template <typename LogEvent, typename ToStringFunc>
+concept DecodeRangeInterface = requires(ToStringFunc func, LogEvent const& log_event) {
+    {
+        func(log_event)
+    } -> std::convertible_to<std::string>;
+};
+
```
junhaoliao, line 65: I appreciate the effort to clean this up as well but since this is not directly related to the feature, shall we revert and address it in another PR?

junhaoliao (dismissed):
lgtm. Please wait for @davemarco 's approval before merging.

junhaoliao:
> From a high level, I noticed we aren't really considering the existing filter. Please feel free to disagree, but I think the function should only consider the filtered events more similiar to `decode_range` behaviour with `use_filter` set. This way it can be more efficient on smaller filtered logs. Secondly, also feel free to disagree. The function should probably return the ActiveLogEventIdx (see [here](https://github.com/y-scope/yscope-log-viewer/blob/main/src/typings/decoders.ts#L46) and [here](https://github.com/y-scope/yscope-log-viewer/blob/07c35584ea2e9ab38293cb015880636157a80fc1/src/utils/data.ts#L81C7-L81C40)) not the log event index. The log event idx isn't terrible, but now we have will need to convert it back to the active log event idx once it is returned to find the correct page. It is probably slightly more efficient to just get the activeLogEventIdx straight from the function. Anyways let me know your thoughts.

Right, if we implement the first point, we will need the second point. Since this is only a performance optimization, it might be fine to leave that into a future PR when we have spare resources.

@Henry8192 if you think this is something quick (less than a morning) to implement, go for it

src/clp_ffi_js/ir/StreamReader.hpp
```diff
@@ -44,6 +46,16 @@ using LogEvents = std::vector<LogEventWithFilterData<LogEvent>>;
  */
 using FilteredLogEventsMap = std::optional<std::vector<size_t>>;
 
+template <typename LogEvent>
```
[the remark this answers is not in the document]
Henry8192: Oh, this was suggested by zhihao, and one of the pros is simplify the require statement in function declaration.

We will populate this concept keyword in the next pr, since junhao wants only the necessary things to be changed

src/clp_ffi_js/ir/StreamReader.cpp
```diff
@@ -145,7 +146,11 @@ EMSCRIPTEN_BINDINGS(ClpStreamReader) {
             )
             .function("filterLogEvents", &clp_ffi_js::ir::StreamReader::filter_log_events)
             .function("deserializeStream", &clp_ffi_js::ir::StreamReader::deserialize_stream)
-            .function("decodeRange", &clp_ffi_js::ir::StreamReader::decode_range);
+            .function("decodeRange", &clp_ffi_js::ir::StreamReader::decode_range)
+            .function(
+                    "getLogEventIndexByTimestamp",
+                    &clp_ffi_js::ir::StreamReader::get_log_event_idx_by_timestamp
```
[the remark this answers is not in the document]
Henry8192: To me their purposes are the same, and plus I'm kinda lazy to change this protocol in the log viewer side XD

src/clp_ffi_js/ir/StructuredIrStreamReader.cpp
```diff
@@ -147,6 +148,15 @@ auto StructuredIrStreamReader::decode_range(size_t begin_idx, size_t end_idx, bo
     );
 }
 
+auto StructuredIrStreamReader::get_log_event_idx_by_timestamp(
+        clp::ir::epoch_time_ms_t const timestamp
+) -> LogEventIdxTsType {
+    return generic_get_log_event_idx_by_timestamp<StructuredLogEvent>(
```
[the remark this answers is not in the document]
Henry8192: emmm you mean `LogEventIdxTsType`?

src/clp_ffi_js/ir/StreamReader.cpp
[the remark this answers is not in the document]
Henry8192: In log viewer the function name was already `getLogEventIdxByTimestamp`.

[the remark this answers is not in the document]
Henry8192: If we want to change the function name, I disagree with `find_log_event_idx_with_nearest_timestamp`. We are not finding the index with nearest timestamp; we are finding the index of the timestamp nearest to our target timestamp.

src/clp_ffi_js/ir/StreamReader.hpp
```diff
@@ -123,6 +135,18 @@ class StreamReader {
...
+     *
+     * @tparam LogEvent
+     * @param timestamp The timestamp to search for, in milliseconds since the Unix epoch.
+     * @return The last index of the log event whose timestamp is smaller than or equal to the
+     * `timestamp`.
+     * @return `0` if all log event timestamps are larger than the target.
+     * @return null if no log event exists in the stream.
+     */
+    [[nodiscard]] virtual auto get_log_event_idx_by_timestamp(clp::ir::epoch_time_ms_t timestamp
+    ) -> LogEventIdxTsType = 0;
```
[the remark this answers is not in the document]
Henry8192: The function description is missing a point that we are finding the **last** index if multiple matches are found.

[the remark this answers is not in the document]
Henry8192: I don't find `timestamp` itself confusing. The member function of getting a log event's timestamp is `.get_timestamp()`; plus `timestamp` is only used once in this function, so I don't think it would cause much confusion.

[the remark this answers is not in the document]
Henry8192: I do think the return statements are better. Let me revise mine and you can take another look.

src/clp_ffi_js/ir/StreamReader.hpp
```diff
@@ -258,6 +285,33 @@ auto StreamReader::generic_filter_log_events(
...
 }
+
+template <typename LogEvent>
+auto StreamReader::generic_get_log_event_idx_by_timestamp(
+        LogEvents<LogEvent> const& log_events,
+        clp::ir::epoch_time_ms_t timestamp
+) -> LogEventIdxTsType {
+    if (log_events.empty()) {
+        return LogEventIdxTsType{emscripten::val::null()};
+    }
```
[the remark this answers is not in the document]
Henry8192: but shouldn't log_events.empty() return true if it is empty? You are giving the wrong suggestion

src/clp_ffi_js/ir/StreamReader.hpp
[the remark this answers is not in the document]
junhaoliao: I agree @davemarco 's suggestion is an improvement

src/clp_ffi_js/ir/StreamReader.cpp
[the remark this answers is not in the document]
Henry8192: offline discussion's conclusion comes to `get_log_event_idx_with_nearest_timestamp`

CMakeLists.txt
```diff
@@ -72,7 +72,11 @@ boost_${CLP_FFI_JS_BOOST_ARCHIVE_VERSION_PART}.tar.gz"
     URL_MD5 "53aeccc3167909ee770e34469f8dd592"
 )
 message(STATUS "Fetching Boost.")
-FetchContent_MakeAvailable(Boost)
+set(FETCHCONTENT_QUIET OFF)
+FetchContent_Declare(
+        Boost
+        URL "${CMAKE_FIND_PACKAGE_REDIRECTS_DIR/boost_1_85_0.tar.gz}"
+)
```
kirkrodrigues, line 79: Is this related to the change?

src/clp_ffi_js/ir/StreamReader.hpp
```diff
@@ -29,6 +30,7 @@ EMSCRIPTEN_DECLARE_VAL_TYPE(ReaderOptions);
 // JS types used as outputs
 EMSCRIPTEN_DECLARE_VAL_TYPE(DecodedResultsTsType);
 EMSCRIPTEN_DECLARE_VAL_TYPE(FilteredLogEventMapTsType);
+EMSCRIPTEN_DECLARE_VAL_TYPE(LogEventIdxTsType);
```
kirkrodrigues, line 33:
```suggestion
// How about `NullableLogEventIdx`?
EMSCRIPTEN_DECLARE_VAL_TYPE(LogEventIdxTsType);
```

src/clp_ffi_js/ir/StreamReader.hpp
```diff
@@ -123,6 +125,17 @@ class StreamReader {
      */
     [[nodiscard]] virtual auto decode_range(size_t begin_idx, size_t end_idx, bool use_filter) const
             -> DecodedResultsTsType = 0;
+    /**
```
kirkrodrigues, line 128:
```suggestion

    /**
```

src/clp_ffi_js/ir/StreamReader.hpp
```diff
@@ -123,6 +125,17 @@ class StreamReader {
...
+     * Finds the log event with the timestamp that's nearest to the `target_ts`.
+     * @param target_ts
+     * @return The index of the log event with:
+     * - the largest timestamp less than or equal to `target_ts`,
+     * - or the index `0` if all timestamps are greater than `target_ts`.
```
kirkrodrigues, line 133:
```suggestion
     * Finds the log event, L, where if we assume:
     *
     * - the collection of log events is sorted in ascending timestamp order;
     * - and we insert a marker log event, M, with timestamp `target_ts` into the collection (if log
     *   events with timestamp `target_ts` already exist in the collection, M should be inserted
     *   after them).
     *
     * L is the event just before M, if M is not the first event in the collection; otherwise L is
     * the event just after M.
     *
     * @param target_ts
     * @return The index of the log event L.
```

I think this method is sufficiently unintuitive that we need to have a more specific description.

src/clp_ffi_js/ir/StreamReader.hpp
```diff
@@ -123,6 +125,17 @@ class StreamReader {
...
+     * @return null if no log event exists in the stream.
+     */
+    [[nodiscard]] virtual auto get_log_event_idx_with_nearest_timestamp(
```
kirkrodrigues, line 136: How about `find_nearest_log_event_by_timestamp`? Similar to Marco's original suggestion. Any ambiguity about the name should be resolved by the docstring.

src/clp_ffi_js/ir/StreamReader.hpp
```diff
@@ -258,6 +285,33 @@ auto StreamReader::generic_filter_log_events(
...
+template <typename LogEvent>
+auto StreamReader::generic_get_log_event_idx_with_nearest_timestamp(
+        LogEvents<LogEvent> const& log_events,
+        clp::ir::epoch_time_ms_t timestamp
+) -> LogEventIdxTsType {
+    if (log_events.empty()) {
+        return LogEventIdxTsType{emscripten::val::null()};
+    }
+
+    auto first_greater_it{std::upper_bound(
```
kirkrodrigues, line 298: std::upper_bound assumes the range is sorted, right? What would happen if the log events aren't in ascending timestamp order? It's technically not impossible and we have seen log files where that's true.

kirkrodrigues, line 298: std::upper_bound assumes the range is sorted, right? What would happen if the log events aren't in ascending timestamp order? It's technically not impossible and we have seen log files where that's true.

kirkrodrigues, line 298:
```suggestion
    // Find the log event whose timestamp is just after `timestamp`.
    auto first_greater_it{std::upper_bound(
```

kirkrodrigues (changes requested):

kirkrodrigues:
Sorry for the delay in reviewing this. Let's keep further discussions in the PR so that we can all come to a consensus more quickly.

CMakeLists.txt
Henry8192: nope. I noticed that just before I go to sleep lol

src/clp_ffi_js/ir/StreamReader.hpp
Henry8192: yeah I was considering this suggested name and the current one. I find this is more logical.

CMakeLists.txt
Henry8192: nope. I noticed that just before I go to sleep lol

src/clp_ffi_js/ir/StreamReader.hpp
Henry8192: Should we also change the name of `FilteredLogEventMapTsType` later? It is also nullable.

kirkrodrigues: I do prefer that, but what do @davemarco / @junhaoliao think?

src/clp_ffi_js/ir/StreamReader.hpp
Henry8192: @junhaoliao iirc, I was told to assume the log events are in accending order. If that's not the case, this whole search function won't work.

If that's not the case, we have to sort all of the log events, cuz' that's the only way to fully garantee all log events are sorted. And we (might?) also need a map to remember their original location.

src/clp_ffi_js/ir/StreamReader.hpp
Henry8192: I personally even prefer making the change here, as it is just a internal type name change and won't affect any front end code.

src/clp_ffi_js/ir/StreamReader.hpp
Henry8192: @kirkrodrigues changed to `find_nearest_log_event_idx_by_timestamp`. I prefer to keep the idx to remind users that we are returning index rather than num (starting from 1). Comment if you disagree.

src/clp_ffi_js/ir/StreamReader.hpp
kirkrodrigues: Oh, isn't `FilteredLogEventIdx` also [defined](https://github.com/y-scope/yscope-log-viewer/blob/main/src/typings/decoders.ts#L36) in the Log Viewer? It's true that the name in clp-ffi-js doesn't influence (from a code execution perspective) the name in the log viewer, but it's probably better for developers if we stay consistent. So if we rename it here, we should probably rename it in the log viewer as well.

src/clp_ffi_js/ir/StreamReader.hpp
kirkrodrigues: Right, idx vs num is a good point. I think I still prefer *not* including `idx` for a few reasons:

- `find_nearest_log_event_idx` makes it sound like we're looking for an index that's nearest to some other index. The `_by_timestamp` part can make this even more confusing since timestamps are also numeric (the user might wonder are we looking for the index that's nearest to the timestamp value?). `find_nearest_log_event_by_timestamp` is a bit more abstract but I think less confusing as a result.
- If the user thinks the method will return a log event, the return value being an integer will correct that misunderstanding at compile time.

It's true that they will have to look at the docstring to determine what the number means. I think the only alternative would be to name the method `find_idx_of_nearest_log_event_by_timestamp`. That's pretty long, but if you think it's better, I'm okay with it.

src/clp_ffi_js/ir/StreamReader.hpp
kirkrodrigues: Since the case with out of order timestamps is rare (as far as we know), we could maybe do something like:

- When decoding the file, keep track of whether the timestamp of two consecutive log events are out of order and if so, set a flag and display a warning to the user that the file contains out of order timestamps and so timestamp-related searches may not function correctly (since two log events with the same timestamp may exist at completely different positions in the log file).
- If that flag is set, we use a brute-force implementation of `find_nearest_log_event_by_timestamp`. Otherwise, we use the implementation you have here.

Henry8192: > * When decoding the file, keep track of whether the timestamp of two consecutive log events are out of order
"keep track of" means we are basically iterating the log events during the decode right?

if that's the case, can we divide the log events up to chunks using these timestamp outliers, perform searches in the proper chunk, and finally iterate those outliers?

Henry8192: duplicated with the previous comment?

kirkrodrigues: Seems like it. Not sure how that happened.

kirkrodrigues: Would that be straightforward to do given that the log events collection changes based on the active filter? If not, then given how rare the case is (as far as we know), I'm not sure the extra code complexity is worth it.

src/clp_ffi_js/ir/StreamReader.hpp
Henry8192: I agree. Let's roll back to your suggestion.

src/clp_ffi_js/ir/StreamReader.hpp
Henry8192: In an offline discussion with @junhaoliao, we reach to an agreement that we can doc the assumption, and neglact this corner case.

By neglacting this corner case, we can save plenty of resource from verifying log events are in chronological order.

src/clp_ffi_js/ir/StreamReader.hpp
```diff
@@ -124,6 +126,24 @@ class StreamReader {
...
+    /**
+     * Finds the log event, L, where if we assume:
+     *
+     * - the collection of log events is sorted in chronological order, or the search won't work;
+     * - and we insert a marker log event, M, with timestamp `target_ts` into the collection (if log
+     *   events with timestamp `target_ts` already exist in the collection, M should be inserted
+     *   after them).
+     *
+     * L is the event just before M, if M is not the first event in the collection; otherwise L is
+     * the event just after M.
```
kirkrodrigues, line 138:
```suggestion
     * - the collection of log events is sorted in chronological order;
     * - and we insert a marker log event, M, with timestamp `target_ts` into the collection (if log
     *   events with timestamp `target_ts` already exist in the collection, M should be inserted
     *   after them).
     *
     * L is the event just before M, if M is not the first event in the collection; otherwise L is
     * the event just after M.
     *
     * NOTE: If the collection of log events isn't in chronological order, this method has undefined
     * behaviour.
```

kirkrodrigues (changes requested):
For the PR title, how about:

```
feat: Add support for finding the log event that's closest to a target timestamp.
```

Henry8192:
@junhaoliao I think this PR is ready to be merged.

src/clp_ffi_js/ir/StreamReader.hpp
```diff
@@ -258,6 +295,34 @@ auto StreamReader::generic_filter_log_events(
...
+    )};
+
+    if (first_greater_it == log_events.begin()) {
+        return NullableLogEventIdx{emscripten::val(0)};
+    }
+
+    auto const first_greater_idx{std::distance(log_events.begin(), first_greater_it)};
+
+    return NullableLogEventIdx{emscripten::val(first_greater_idx - 1)};
+}
```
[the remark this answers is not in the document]
Henry8192: We have that assumption.

kirkrodrigues (approved):

PR #45, opened 2024-12-27
build(cmake): Extract CMake project name and version from package.json.

junhaoliao:
# Description

1. Extract CMake project name and version from `package.json`.

# Validation performed

## Positive test
1. Ran `task` in the project root.
2. Under the project root, `cd build/clp-ffi-js`.
3. Ran `cmake -LAH | grep -B 1 CLP_FFI_JS` and saw below in the output:
    ```
    ...
    // The name of the project parsed from `package.json`.
    CLP_FFI_JS_PROJECT_NAME:STRING=clp-ffi-js
    --
    // The version of the project parsed from `package.json`.
    CLP_FFI_JS_VERSION:STRING=0.3.1
    ```
4. Compare that with the one in `package.json` and found no discrepancy.

## Negative test
1. Removed line `"version": "0.3.1",` from `package.json`.
2. Repeat above positive test steps 1-3 and saw
    ```
    ...
    // The name of the project parsed from `package.json`.
    CLP_FFI_JS_PROJECT_NAME:STRING=clp-ffi-js
    --
    // Use a placeholder version.
    CLP_FFI_JS_VERSION:STRING=0.0.0
    ```

## Summary by CodeRabbit

- **New Features**
	- Project name and version are now dynamically extracted from `package.json`, enhancing configurability.
- **Bug Fixes**
	- Improved handling of project name and version defaults if not specified in `package.json`.
- **Chores**
	- Updated build configurations to ensure correct toolchain usage for Emscripten builds.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,11 +11,11 @@ endif()
 
 # Extract the project name & version from package.json
 if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/package.json")
-    file(READ "${CMAKE_CURRENT_SOURCE_DIR}/package.json" PACKAGE_JSON_CONTENT)
+    file(READ "${CMAKE_CURRENT_SOURCE_DIR}/package.json" CLP_FFI_JS_PACKAGE_JSON_CONTENT)
 else()
     message(FATAL_ERROR "`package.json` not found in ${CMAKE_CURRENT_SOURCE_DIR}")
 endif()
-if("${PACKAGE_JSON_CONTENT}" MATCHES "\"name\":[ ]*\"([^\"]+)\"")
+if("${CLP_FFI_JS_PACKAGE_JSON_CONTENT}" MATCHES "\"name\":[ ]*\"([^\"]+)\"")
     set(CLP_FFI_JS_PROJECT_NAME
             "${CMAKE_MATCH_1}"
             CACHE STRING
@@ -25,7 +25,7 @@ if("${PACKAGE_JSON_CONTENT}" MATCHES "\"name\":[ ]*\"([^\"]+)\"")
 else()
     set(CLP_FFI_JS_PROJECT_NAME "clp-ffi-js" CACHE STRING "Use a placeholder project name." FORCE)
 endif()
-if("${PACKAGE_JSON_CONTENT}" MATCHES "\"version\":[ ]*\"([^\"]+)\"")
+if("${CLP_FFI_JS_PACKAGE_JSON_CONTENT}" MATCHES "\"version\":[ ]*\"([^\"]+)\"")
     set(CLP_FFI_JS_VERSION
         "${CMAKE_MATCH_1}"
         CACHE STRING
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,12 +9,35 @@ Emscripten.cmake"
     )
 endif()
 
+# Extract the project name & version from package.json
+file(READ "${CMAKE_CURRENT_SOURCE_DIR}/package.json" PACKAGE_JSON_CONTENT)
+if(${PACKAGE_JSON_CONTENT} MATCHES "\"name\":[ ]*\"([^\"]+)\"")
+    set(CLP_FFI_JS_PROJECT_NAME
+            ${CMAKE_MATCH_1}
+            CACHE STRING
+            "The name of the project parsed from `package.json`."
+            FORCE
+    )
+else()
+    set(CLP_FFI_JS_PROJECT_NAME "clp-ffi-js" CACHE STRING "Use a placeholder project name." FORCE)
+endif()
+if(${PACKAGE_JSON_CONTENT} MATCHES "\"version\":[ ]*\"([^\"]+)\"")
+    set(CLP_FFI_JS_VERSION
+        ${CMAKE_MATCH_1}
+        CACHE STRING
+        "The version of the project parsed from `package.json`."
+        FORCE
+    )
+else()
+    set(CLP_FFI_JS_VERSION "0.0.0" CACHE STRING "Use a placeholder version." FORCE)
+endif()
+
 project(
-    clp-ffi-js
+    ${CLP_FFI_JS_PROJECT_NAME}
     LANGUAGES
         C
         CXX
-    VERSION 0.3.1
+    VERSION "${CLP_FFI_JS_VERSION}"
 )
 
 # Enable exporting compile commands
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,7 +37,7 @@ else()
 endif()
 
 project(
-    ${CLP_FFI_JS_PROJECT_NAME}
+    "${CLP_FFI_JS_PROJECT_NAME}"
     LANGUAGES
         C
         CXX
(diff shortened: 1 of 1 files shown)


junhaoliao:
We could also have used [string(JSON ...)](https://cmake.org/cmake/help/latest/command/string.html#json) but that is
> Added in version 3.19.

We're currently enforcing `cmake_minimum_required(VERSION 3.16)` and upgrading that would require developer who use Ubuntu Focal (20.04) to install `cmake` from non-Ubuntu-apt sources, which is non-ideal.

CMakeLists.txt
```diff
@@ -9,12 +9,39 @@ Emscripten.cmake"
...
+        "The version of the project parsed from `package.json`."
+        FORCE
+    )
+else()
+    set(CLP_FFI_JS_VERSION "0.0.0" CACHE STRING "Use a placeholder version." FORCE)
+endif()
+
 project(
-    clp-ffi-js
+    ${CLP_FFI_JS_PROJECT_NAME}
```
junhaoliao, line 40:
```suggestion
    "${CLP_FFI_JS_PROJECT_NAME}"
```

CMakeLists.txt
```diff
@@ -9,12 +9,39 @@ Emscripten.cmake"
     )
 endif()
 
+# Extract the project name & version from package.json
+if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/package.json")
+    file(READ "${CMAKE_CURRENT_SOURCE_DIR}/package.json" PACKAGE_JSON_CONTENT)
```
kirkrodrigues, line 14: How about `CLP_FFI_JS_PACKAGE_JSON_CONTENT`? (To follow our convention of prefixing local variables with the project name.)

CMakeLists.txt
```diff
@@ -9,12 +9,39 @@ Emscripten.cmake"
...
+else()
+    message(FATAL_ERROR "`package.json` not found in ${CMAKE_CURRENT_SOURCE_DIR}")
+endif()
+if(${PACKAGE_JSON_CONTENT} MATCHES "\"name\":[ ]*\"([^\"]+)\"")
```
kirkrodrigues, line 18:
```suggestion
if("${PACKAGE_JSON_CONTENT}" MATCHES "\"name\":[ ]*\"([^\"]+)\"")
```
To avoid [double-variable expansion](https://cmake.org/cmake/help/latest/command/if.html#variable-expansion).

CMakeLists.txt
```diff
@@ -9,12 +9,39 @@ Emscripten.cmake"
...
+    set(CLP_FFI_JS_PROJECT_NAME
+            ${CMAKE_MATCH_1}
+            CACHE STRING
+            "The name of the project parsed from `package.json`."
+            FORCE
```
kirkrodrigues, line 23: Why do we need/want to make these cache variables?

CMakeLists.txt
```diff
@@ -9,12 +9,39 @@ Emscripten.cmake"
...
 
+# Extract the project name & version from package.json
+if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/package.json")
+    file(READ "${CMAKE_CURRENT_SOURCE_DIR}/package.json" PACKAGE_JSON_CONTENT)
+else()
+    message(FATAL_ERROR "`package.json` not found in ${CMAKE_CURRENT_SOURCE_DIR}")
+endif()
+if(${PACKAGE_JSON_CONTENT} MATCHES "\"name\":[ ]*\"([^\"]+)\"")
+    set(CLP_FFI_JS_PROJECT_NAME
+            ${CMAKE_MATCH_1}
```
kirkrodrigues, line 20:
```suggestion
            "${CMAKE_MATCH_1}"
```

CMakeLists.txt
```diff
@@ -9,12 +9,39 @@ Emscripten.cmake"
...
+            CACHE STRING
+            "The name of the project parsed from `package.json`."
+            FORCE
+    )
+else()
+    set(CLP_FFI_JS_PROJECT_NAME "clp-ffi-js" CACHE STRING "Use a placeholder project name." FORCE)
+endif()
+if(${PACKAGE_JSON_CONTENT} MATCHES "\"version\":[ ]*\"([^\"]+)\"")
+    set(CLP_FFI_JS_VERSION
+        ${CMAKE_MATCH_1}
```
kirkrodrigues, line 30:
```suggestion
        "${CMAKE_MATCH_1}"
```

CMakeLists.txt
```diff
@@ -9,12 +9,39 @@ Emscripten.cmake"
...
+    set(CLP_FFI_JS_PROJECT_NAME
+            ${CMAKE_MATCH_1}
+            CACHE STRING
+            "The name of the project parsed from `package.json`."
+            FORCE
+    )
+else()
+    set(CLP_FFI_JS_PROJECT_NAME "clp-ffi-js" CACHE STRING "Use a placeholder project name." FORCE)
+endif()
+if(${PACKAGE_JSON_CONTENT} MATCHES "\"version\":[ ]*\"([^\"]+)\"")
```
kirkrodrigues, line 28:
```suggestion
if("${PACKAGE_JSON_CONTENT}" MATCHES "\"version\":[ ]*\"([^\"]+)\"")
```

kirkrodrigues (changes requested):

CMakeLists.txt
junhaoliao: my understanding of CMake cache variables is that they are like configuration variables that can be overridden by `-DXXX` passed to the `cmake` command. 

with that said, realistically, in this project, those variables are not getting referenced anywhere else, so hardcoding those (project_name=clp-ffi-js; version=0.0.0) may also work just fine. (with that said, those variables might come in handy if we ever use Doxygen to generate docs for the C++ code / add any other tools that might need such project metadata. ) what do you think?

junhaoliao: and setting cache variables allow docstrings (though it's not like we can't write a comment instead

i guess they don't strictly need to be cache variables

kirkrodrigues: > but yeah why don't we directly change the metadata in package.json and let cmake sync from it? lol

Yep.

> with that said, realistically, in this project, those variables are not getting referenced anywhere else, so hardcoding those (project_name=clp-ffi-js; version=0.0.0) may also work just fine. (with that said, those variables might come in handy if we ever use Doxygen to generate docs for the C++ code / add any other tools that might need such project metadata. ) what do you think?

I think this point is kind of unrelated to the above and sounds more like it's questioning the motivation of the PR? Anyway, it's not a huge amount of effort to parse the project name and version out of package.json (and you've already done it), so I think we should keep it.

> i guess they don't strictly need to be cache variables

Yep.

junhaoliao: > I think this point is kind of unrelated to the above and sounds more like it's questioning the motivation of the PR?

yes

> I think we should keep it.

sure. sounds good

kirkrodrigues (approved):
For the PR title, how about:

```
build(cmake): Extract CMake project name and version from package.json.
```

It's subjective, but I think `build` is a more appropriate type since this change feels like more than just a chore.

PR #61, opened 2025-02-21
feat!: Remove timestamp formatting from unstructured log messages.

AVMatthews:
# Description

Remove all references to m_ts_pattern from the UnstructuredIRStreamReader. We will no longer be formatting these timestamps here and will be allowing the log-viewer to handle formatting it using a custom DayJS format string from the user.

# Checklist

  breaking change.

# Validation performed

Made sure the timestamp was not inserted into the message and that the updated files ran successfully with the log-viewer.

## Summary by CodeRabbit

- **Refactor**
  - Simplified log event processing by removing the inclusion of a formatted timestamp in log messages.
  - Streamlined component initialization by eliminating dependencies related to timestamp handling.

### DIFF
--- a/src/clp_ffi_js/ir/UnstructuredIrStreamReader.cpp
+++ b/src/clp_ffi_js/ir/UnstructuredIrStreamReader.cpp
@@ -160,9 +160,10 @@ auto UnstructuredIrStreamReader::decode_range(size_t begin_idx, size_t end_idx, 
 UnstructuredIrStreamReader::UnstructuredIrStreamReader(
         StreamReaderDataContext<UnstructuredIrDeserializer>&& stream_reader_data_context
 )
-        : m_stream_reader_data_context{std::make_unique<
-                  StreamReaderDataContext<UnstructuredIrDeserializer>>(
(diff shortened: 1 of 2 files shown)


src/clp_ffi_js/ir/UnstructuredIrStreamReader.cpp
```diff
@@ -144,7 +144,6 @@ auto UnstructuredIrStreamReader::decode_range(size_t begin_idx, size_t end_idx,
             };
         }
         message = parsed.value();
-        m_ts_pattern.insert_formatted_timestamp(log_event.get_timestamp(), message);
         return message;
```
junhaoliao, line 147:
```suggestion
        return parsed.value();
```
Then we can delete the `std::string message;` declaration and the value assignment above.

junhaoliao (changes requested):

junhaoliao:
edited the PR title directly to include a period.

junhaoliao:
edited the PR title again to indicate this is a breaking change.

PR #76, opened 2025-04-28
WIP feat: Implement StreamWriter and StructuredIrStreamWriter for IR data streaming (based on #74).

zzxthehappiest:
# Description

Implemented the desiredSize/ready and abort on top of #74 

**Done:**
+ Add and bind three new functions in StreamWriter class:
    + `getLastWritePromise`: return a `Promise` for the last write operation. Since the `clp::WriterInterface::write` returns void, to make the `WritableStream` can correctly backpressure (as the [example here](https://streams.spec.whatwg.org/#example-manual-write-with-backpressure)), we need `await writer.ready` (which is an async operation) in `WebStreamWriter::write`, so it also needs to return a `Promise` so in JS code the `Promise` produced by `await writer.ready` can be resolved correctly, avoid memory leak after main function exits (if not doing so, there will be a memory access exception).
    + `desiredSize`: return the desiredSize property of the `WritableStream`, it is a sync function.
    + `abort`: call the `abort(reason)` defined in the `underlying sink` in the`WritableStream`.
+ Add a helper function in `WebStreamWriter`:
    + `await_ready`: get the readyPromise property of the `WritableStream`, so that we can add callback to make the `WebStreamWriter::write` correctly write the content after backpressure is finished.
+ Rewrite `WebStreamWriter::write`:
    + Since the previous version is a pure sync so it is not able to use desiredSize/ready backpressure mechnism of `WritableStream`, so I make it a "fake"-async function.
+ Refactor the unit test:
    + Modify the `CMakeLists.txt` now it copies the generated `.js` files to the unit test directly, so we don't need to hard code it as `cmake-build-release` which is only used by Clion.
    + Add the unit test for backpressure and abort.
    + Add the unit test for `ClpFfiJs-worker`, but haven't successfully run it for unknown reason.
    
** Questions:**
+ For the following FIXME I am not sure what do they mean exactly:
    + In `StreamWriter::write` ([link](https://github.com/y-scope/clp-ffi-js/pull/74/files#diff-c05bee3821b6c895c7b2b84e9bddaddfd8d921cb6f36b3003550446673bb1cdbR37))
    + In `StreamWriter::flush` ([link](https://github.com/y-scope/clp-ffi-js/pull/74/files#diff-c05bee3821b6c895c7b2b84e9bddaddfd8d921cb6f36b3003550446673bb1cdbR43)). For the `flush` I checked the `Compressor.cpp`, it seems we never use `m_compressed_stream_writer.flush()` so I am not sure do we really need to implement this in the `WebStreamWriter` as well as modify the current`StructuredIrStreamWriter::flush`.
    + In `StructuredIrStreamWriter::write` ([link](https://github.com/y-scope/clp-ffi-js/pull/74/files#diff-1fcf88983ce98537c503b0b02c007c97a85534fece1ebe98d8c9d8d4611915b2R105))
    + In `StructuredIrStreamWriter::close` ([link](https://github.com/y-scope/clp-ffi-js/pull/74/files#diff-1fcf88983ce98537c503b0b02c007c97a85534fece1ebe98d8c9d8d4611915b2R134))
+ For the following TODO I am not sure what do they mean exactly:
    + In `StreamWriter::write` ([link](https://github.com/y-scope/clp-ffi-js/pull/74/files#diff-c05bee3821b6c895c7b2b84e9bddaddfd8d921cb6f36b3003550446673bb1cdbR57))
    + In the unit test ([link](https://github.com/y-scope/clp-ffi-js/pull/74/files#diff-2b297e381f80175634ab045ed453022ac6eeb7fd7ce41357e9d5ec33222490deR36))
+ To support the backpressure we need to support async; to support async we need to remove the `-fwasm-exceptions` compile flag and add `-sASYNCIFY=1` flag, otherwise there is compile error says the async is not supported. However, there is an apparent performance cost of it, the 1000000 write operations takes around 0.7s slower than before (2.1s -> 2.8s). It seems these two compile flags have not been supported at the same time yet in the `emsk` framework. But from the stream standard, `WritableStreamDefaultWriter`'s `abort`, `close` and `write` are all async (they all return `Promise`, [link](https://streams.spec.whatwg.org/#default-writer-class-definition)), as well as the same functions in the `underlying sink` ([link](https://streams.spec.whatwg.org/#underlying-sink-api)). So I think it is worth to trade off the cost to support the async feature and we can potentially gain the performance back in other places (e.g., better caching policy when `write`).

# Checklist

* [ ] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [ ] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [ ] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

Run the unit test `test-node.mjs`.


junhaoliao:
Sorry for this late response. Thanks for helping and please see below for my answers:

> + For the following FIXME I am not sure what do they mean exactly:
>    + In `StreamWriter::write` ([link](https://github.com/y-scope/clp-ffi-js/pull/74/files#diff-c05bee3821b6c895c7b2b84e9bddaddfd8d921cb6f36b3003550446673bb1cdbR37))

Sorry for the confusion. I meant to say instead of implementing a `write()` method which accepts some `chunk` as bytes, we may want a separate interface (e.g., `appendLogEvent(event: object)` ) to accept some argument of type `object` which contains kv-pairs. In `appendLogEvent`, after serializing the object into msgpack bytes, `appendLogEvent` can call `write(chunk)` where the chunk will be the msgpack bytes. 

However, during prototyping, as we may find via the commit history, it was found that serializing any JS-object in C++ is quite expensive (~4x slower) because there will be a lot of cross-VM communications. Instead, I decided to serialize the object -> msgpack in JS then pass the msgpack bytes into C++ code. Luckily, it can be easily achieved via the `msgpackr` library as we can see in the latest `test.mjs`.

Nevertheless, please ignore this action item for now. I am thinking of a different interface due to some limitations in the current implementation. Please see below for the explanation.

---

In the prototype, I was hoping to implement the CLP IR writer with some common / standard interface in JavaScript, so it may work better with other (JS-standard / 3rd-party) reader / writer / transform streams. However, after conducting some throughout research this week, I found `WritableStreamDefaultWriter`'s extensibility is quite limited - it seems no web code developer is supposed to directly call the constructor of a `WritableStreamDefaultWriter` because such instances are only supposed to be created via `getWriter()` on a `WritableStream` instance. For flush control, the standard does provide a `WritableStreamDefaultController` class, but any control seems to only happen in the `underlyingSink`'s `write()` method, which isn't directly achievable by simply extending `WritableStreamDefaultWriter`. Even worse, I also couldn't find any example about extending the `WritableStreamDefaultWriter` class in any official docs / online. 

I believe some changes are needed in the interface. Here are some approaches I am picturing:
1. In addition to extending `WritableStreamDefaultWriter` as `StructuredIrStreamWriter`, we can also extend `WritableStream` as `StructuredIrWritableStream`. In the extended `WritableStream` class, `getWriter()` returns a `StructuredIrStreamWriter` instance for `clp-ffi-js` library users to write msgpack bytes as chunks. For flush control on write, `StructuredIrWritableStream` will execute its own logic before calling the underlying sink's `write` method. In addition, for explicit flush control by library users, `StructuredIrWritableStream` will also provide a `flush()` method.
   Below is some pseudo code for illustration:
    ```js
    class StructuredIrWritableStream extends WritableStream {
        constructor(underlyingSink, strategy) {
            this.#writeToSink = underlyingSink.write
    
            super({
                ...underlyingSink,
                write: (chunk) => {
                    // our own flush-on-write logic
                    if (shouldFlush) {
                        underlyingSink.write(chunk)
                    }
                }
            }, strategy)
        }
    
        flush() {
            this.#writeToSink(chunk)
        }
    
        getWriter() {
            return StructuredIrStreamWriter(stream, DEFAULT_WRITER_OPTIONS)
        }
    
        getWriterWithOptions(options) {
            return StructuredIrStreamWriter(stream, options)
        }
    }
    ```
2. OR: Implement a custom `StructuredIrStreamWriter` instead of extending `WritableStreamDefaultWriter`, so we're not constrained by the lifecycle requirements of the `WritableStream` utilities.
3. OR: In addition to (2.) where we implement our own `StructuredIrStreamWriter`, provide a method `createUnderlyingSink`, which returns a `UnderlyingSink` dictionary to be used in `WritableStream`'s constructor. Essentially, in this approach we are providing an "underlying sink" instead of a writer to our library users.

cc @LinZhihao-723 @davidlion @kirkrodrigues @hoophalab for comments, who are experts with writers in Python / Golang.

I'll respond to the rest of the questions in a separate thread to avoid spamming the experts.

junhaoliao:
@zzxthehappiest @hoophalab 

@davidlion reviewed this and we discussed offline -
The primary reason for implementing existing JavaScript interfaces is to enable integration with other stream types, such as CompressionStream. (What @hoophalab proposed will work well for this purpose - CompressionStream extends TranformStream.) However, we should note CompressionStream doesn't yet support Zstandard (zstd) compression, the de-facto standard algorithm used in CLP IR formats.
Besides that, extending the WritableStream class specifically for logging purposes offers limited benefits, given its lack of adoption in this context. Instead, as @davidlion proposed, we want to follow the interface patterns already established in our `clp-ffi-py` and `clp-ffi-go` libraries, focusing on ensuring compatibility with popular JavaScript logging frameworks like Pino and Winston rather than adhering to the WritableStream lifecycle.

zzxthehappiest:
> @zzxthehappiest Sorry for this late response. Thanks for helping and please see below for my answers:
> 
> > * For the following FIXME I am not sure what do they mean exactly:
> >   
> >   * In `StreamWriter::write` ([link](https://github.com/y-scope/clp-ffi-js/pull/74/files#diff-c05bee3821b6c895c7b2b84e9bddaddfd8d921cb6f36b3003550446673bb1cdbR37))
> 
> Sorry for the confusion. I meant to say instead of implementing a `write()` method which accepts some `chunk` as bytes, we may want a separate interface (e.g., `appendLogEvent(event: object)` ) to accept some argument of type `object` which contains kv-pairs. In `appendLogEvent`, after serializing the object into msgpack bytes, `appendLogEvent` can call `write(chunk)` where the chunk will be the msgpack bytes.
> 
> However, during prototyping, as we may find via the commit history, it was found that serializing any JS-object in C++ is quite expensive (~4x slower) because there will be a lot of cross-VM communications. Instead, I decided to serialize the object -> msgpack in JS then pass the msgpack bytes into C++ code. Luckily, it can be easily achieved via the `msgpackr` library as we can see in the latest `test.mjs`.
> 
> Nevertheless, please ignore this action item for now. I am thinking of a different interface due to some limitations in the current implementation. Please see below for the explanation.
> 
> In the prototype, I was hoping to implement the CLP IR writer with some common / standard interface in JavaScript, so it may work better with other (JS-standard / 3rd-party) reader / writer / transform streams. However, after conducting some throughout research this week, I found `WritableStreamDefaultWriter`'s extensibility is quite limited - it seems no web code developer is supposed to directly call the constructor of a `WritableStreamDefaultWriter` because such instances are only supposed to be created via `getWriter()` on a `WritableStream` instance. For flush control, the standard does provide a `WritableStreamDefaultController` class, but any control seems to only happen in the `underlyingSink`'s `write()` method, which isn't directly achievable by simply extending `WritableStreamDefaultWriter`. Even worse, I also couldn't find any example about extending the `WritableStreamDefaultWriter` class in any official docs / online.
> 
> I believe some changes are needed in the interface. Here are some approaches I am picturing:
> 
> 1. In addition to extending `WritableStreamDefaultWriter` as `StructuredIrStreamWriter`, we can also extend `WritableStream` as `StructuredIrWritableStream`. In the extended `WritableStream` class, `getWriter()` returns a `StructuredIrStreamWriter` instance for `clp-ffi-js` library users to write msgpack bytes as chunks. For flush control on write, `StructuredIrWritableStream` will execute its own logic before calling the underlying sink's `write` method. In addition, for explicit flush control by library users, `StructuredIrWritableStream` will also provide a `flush()` method.
>    Below is some pseudo code for illustration:
>    ```js
>    class StructuredIrWritableStream extends WritableStream {
>        constructor(underlyingSink, strategy) {
>            this.#writeToSink = underlyingSink.write
>    
>            super({
>                ...underlyingSink,
>                write: (chunk) => {
>                    // our own flush-on-write logic
>                    if (shouldFlush) {
>                        underlyingSink.write(chunk)
>                    }
>                }
>            }, strategy)
>        }
>    
>        flush() {
>            this.#writeToSink(chunk)
>        }
>    
>        getWriter() {
>            return StructuredIrStreamWriter(stream, DEFAULT_WRITER_OPTIONS)
>        }
>    
>        getWriterWithOptions(options) {
>            return StructuredIrStreamWriter(stream, options)
>        }
>    }
>    ```
> 2. OR: Implement a custom `StructuredIrStreamWriter` instead of extending `WritableStreamDefaultWriter`, so we're not constrained by the lifecycle requirements of the `WritableStream` utilities.
> 3. OR: In addition to (2.) where we implement our own `StructuredIrStreamWriter`, provide a method `createUnderlyingSink`, which returns a `UnderlyingSink` dictionary to be used in `WritableStream`'s constructor. Essentially, in this approach we are providing an "underlying sink" instead of a writer to our library users.
> 
> cc @LinZhihao-723 @davidlion @kirkrodrigues @hoophalab for comments, who are experts with writers in Python / Golang.
> 
> I'll respond to the rest of the questions in a separate thread to avoid spamming the experts.

If I understand correctly, we are going to try the second option with what @hoophalab proposed?

PR #81, opened 2025-05-26
chore: Update linting requirements and tasks for improved code formatting.

junhaoliao:
# Description

1. Update submodule `yscope-dev-utils` version to y-scope/yscope-dev-utils@76aecbd6d0db6e2078005431946b5c2e716c4291
2. Update references to `yscope-dev-utils` utilities in both `Taskfile.yml` and `lint-tasks.yml`.
3. Update versions in `linting-requirements.txt` using https://github.com/y-scope/clp/blob/8d008abe40ca1ae4531d05b4dbd9085ff0ce53ed/lint-requirements.txt as a reference.
4. Apply latest formatting and linting configurations by running the updated task.

# Checklist

  breaking change.

# Validation performed

TBA

### DIFF
--- a/lint-tasks.yml
+++ b/lint-tasks.yml
@@ -1,127 +1,154 @@
 version: "3"
 
 vars:
+  G_LINT_CLANG_TIDY_DIR: "{{.G_BUILD_DIR}}/lint-clang-tidy"
+  G_LINT_VENV_DIR: "{{.G_BUILD_DIR}}/lint-venv"
   G_SRC_DIR: "{{.ROOT_DIR}}/src"
   G_SRC_CLP_FFI_JS_DIR: "{{.G_SRC_DIR}}/clp_ffi_js"
-  G_LINT_VENV_DIR: "{{.G_BUILD_DIR}}/lint-venv"
 
 tasks:
   check:
     cmds:
-      - task: "cpp-check"
-      - task: "yml-check"
+      - task: "check-cpp-full"
+      - task: "check-yaml"
+
+  check-cpp-diff:
+    cmds:
+      - task: "check-cpp-format"
+      - task: "check-cpp-static-diff"
+
+  check-cpp-full:
+    cmds:
+      - task: "check-cpp-format"
+      - task: "check-cpp-static-full"
+
+  check-no-cpp:
+    cmds:
+      - task: "check-yaml"
 
   fix:
     cmds:
-      - task: "cpp-fix"
-      - task: "yml-fix"
+      - task: "fix-cpp-full"
+      - task: "fix-yaml"
 
-  cpp-configs:
-    cmd: "{{.ROOT_DIR}}/tools/yscope-dev-utils/lint-configs/symlink-cpp-lint-configs.sh"
-
-  cpp-check:
+  fix-cpp-diff:
     cmds:
-      - task: "cpp-format-check"
-      - task: "cpp-static-check"
+      - task: "fix-cpp-format"
+      - task: "fix-cpp-static-diff"
 
-  cpp-fix:
+  fix-cpp-full:
     cmds:
-      - task: "cpp-format-fix"
-      - task: "cpp-static-fix"
+      - task: "fix-cpp-format"
+      - task: "fix-cpp-static-full"
 
-  cpp-format-check:
+  fix-no-cpp:
+    cmds:
+      - task: "fix-yaml"
+
+  check-cpp-format:
     sources: &cpp_format_source_files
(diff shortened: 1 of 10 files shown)


davidlion (changes requested):
Lgtm other than https://github.com/y-scope/clp-ffi-js/pull/81/files#r2106490732.

Maybe it is worth using a multi-line commit message since the PR does a handful of things. Also, it is probably worth adding a line that some tasks like clang format/tidy were changed to use dev-utils' version from the find+xargs/internal version.

junhaoliao:
@davidlion 
> Maybe it is worth using a multi-line commit message since the PR does a handful of things. Also, it is probably worth adding a line that some tasks like clang format/tidy were changed to use dev-utils' version from the find+xargs/internal version.

I updated the PR title. what do you think about it?

also, here are the rest of the multi-line commit message I plan to write:
```
- Update paths and task names accordingly in Taskfile.yml and lint-tasks.yml.
- Replace internal find + xargs usage for clang-format/clang-tidy with utility tasks from yscope-dev-utils.
- Bump clang-format and clang-tidy to version 20.1.
```

feel free to make any suggestions

lint-tasks.yml
```diff
@@ -1,127 +1,134 @@
...
+      - "venv"
+    cmds:
+      - task: ":utils:cpp-lint:clang-tidy-find"
+        vars:
+          FLAGS: *clang_tidy_flags
+          INCLUDE_PATTERNS:
+            - "{{.G_SRC_CLP_FFI_JS_DIR}}/**"
+          OUTPUT_DIR: "{{.G_LINT_CLANG_TIDY_DIR}}"
+          ROOT_PATHS:
+            - "{{.G_SRC_CLP_FFI_JS_DIR}}/**/*"
```
davidlion, line 110: Sorry I should have spotted this in the first review. This suggestion _should_ work, but please double check just in case.

```suggestion
          OUTPUT_DIR: "{{.G_LINT_CLANG_TIDY_DIR}}"
          ROOT_PATHS:
            - "{{.G_SRC_CLP_FFI_JS_DIR}}"
```

relevant doc-string snippets:
> * ROOT_PATHS Paths recursively searched for files to run on.
> * [INCLUDE_PATTERNS] Wildcard patterns where any matching file paths (relative
   to `ROOT_PATHS`) will be included. By default, all paths match.
   
explanations:
* Adding child paths of `G_SRC_CLP_FFI_JS_DIR` to `ROOT_PATHS` is unnecessary because `find` recursively searches each path in `ROOT_PATHS`.
* Having `INCLUDE_PATTERNS` match anything under `G_SRC_CLP_FFI_JS_DIR` is unnecessary as by default `INCLUDE_PATTERNS` will match everything under the `ROOT_PATHS`.

lint-tasks.yml
```diff
@@ -1,127 +1,134 @@
...
+      - "tools/yscope-dev-utils/exports/lint-configs/.clang-format"
+    deps: ["cpp-lint-configs", "venv"]
     cmds:
-      - task: "clang-format"
+      - task: ":utils:cpp-lint:clang-format"
         vars:
-          FLAGS: "--dry-run"
+          FLAGS: ["--dry-run"]
+          INCLUDE_FILENAME_PATTERNS: ["*.cpp", "*.h", "*.hpp", "*.inc"]
+          ROOT_PATHS: *cpp_format_source_files
```
davidlion, line 60: Optional: technically this will make `find` try to (recursively) search from each individual file matching the `sources` list. I don't really see any correctness problem with it, but it is a bit weird logically.
The same is true in `fix-cpp-format`.

```suggestion
          ROOT_PATHS:
            - "{{.G_SRC_CLP_FFI_JS_DIR}}"
```

lint-tasks.yml
```diff
@@ -1,127 +1,134 @@
...
-
-      # Other sources
-      - "{{.G_CLP_FFI_JS_BUILD_DIR}}/compile_commands.json"
-      - "{{.ROOT_DIR}}/Taskfile.yml"
-      - "{{.TASKFILE}}"
-      - "tools/yscope-dev-utils/lint-configs/.clang-tidy"
-    deps: [":config-cmake-project", "cpp-configs", "venv"]
+          FLAGS: ["-i"]
+          INCLUDE_FILENAME_PATTERNS: ["*.cpp", "*.h", "*.hpp", "*.inc"]
+          ROOT_PATHS: *cpp_format_source_files
```
junhaoliao, line 71:
```suggestion
          ROOT_PATHS:
            - "{{.G_SRC_CLP_FFI_JS_DIR}}"
```

davidlion (approved):
I updated the title/description btw.

PR #99, opened 2025-06-25
feat: Add support for parsing UTC offset in seconds per log event and returning it in decode results (resolves #79).

### LINKED ISSUE #79: Return time zone information for Unstructured IR log events.
### Request

as the title suggests

### Possible implementation

TBA

hoophalab:
# Description

Add utcOffset field to decoded log event results in StreamReader

1. **LogEventWithFilterData.hpp**  
   - Added `m_utc_offset` and updated constructor/getters to include UTC offset.

2. **StreamReader.cpp**
   - Updated `ReaderOptions` and `DecodedResultsTsType` to include `utcOffset`.

3. **StreamReader.hpp**
   - Modified `generic_decode_range` to add `utcOffset` to JS output.

4. **StructuredIrUnitHandler.cpp**
   - Implemented `get_utc_offset` to include UTC offset from key value pairs.

5. **UnstructuredIrStreamReader.cpp**
   - Updated `deserialize_stream` to include UTC offset from `UtcOffsetChange` payload.

# Checklist

  breaking change.

# Validation performed

Run the new `decodeRange` in this [branch](https://github.com/hoophalab/yscope-log-viewer/tree/test-decode-range) of log viewer.

## Summary by CodeRabbit

- **New Features**
  - Decoded log event results now include a `utcOffset` property, showing each event's timezone offset in minutes relative to UTC.
  - Log event input options expanded to support specifying a UTC offset key for more precise time zone handling.
  - Structured and unstructured log readers updated to process and expose UTC offset information alongside existing log details.

### DIFF
--- a/src/clp_ffi_js/ir/StreamReader.cpp
+++ b/src/clp_ffi_js/ir/StreamReader.cpp
@@ -73,8 +73,7 @@ EMSCRIPTEN_BINDINGS(ClpStreamReader) {
             .value("UNSTRUCTURED", clp_ffi_js::ir::StreamType::Unstructured);
     emscripten::register_type<clp_ffi_js::ir::DecodedResultsTsType>(
             "Array<{logEventNum: number, logLevel: number, message: string, timestamp: bigint, "
-            "utcOffsetMinutes: bigint}> "
-            "| null"
+            "utcOffset: bigint}> | null"
     );
     emscripten::register_type<clp_ffi_js::ir::FilteredLogEventMapTsType>("number[] | null");
     emscripten::register_type<clp_ffi_js::ir::NullableLogEventIdx>("number | null");
--- a/src/clp_ffi_js/ir/StructuredIrUnitHandler.cpp
+++ b/src/clp_ffi_js/ir/StructuredIrUnitHandler.cpp
@@ -329,6 +329,10 @@ auto StructuredIrUnitHandler::get_utc_offset(StructuredLogEvent const& log_event
         );
         return cDefaultUtcOffset;
     }
-    return UtcOffset{utc_offset_value.get_immutable_view<clp::ffi::value_int_t>()};
+    // TODO: Support parsing of other time units, e.g., minutes, hours, etc..
+    auto const utc_offset_seconds{
+            std::chrono::seconds{utc_offset_value.get_immutable_view<clp::ffi::value_int_t>()}
+    };
+    return std::chrono::duration_cast<UtcOffset>(utc_offset_seconds);
 }
 }  // namespace clp_ffi_js::ir


src/clp_ffi_js/ir/LogEventWithFilterData.hpp
```diff
@@ -30,11 +31,13 @@ class LogEventWithFilterData {
     LogEventWithFilterData(
             LogEvent log_event,
             LogLevel log_level,
-            clp::ir::epoch_time_ms_t timestamp
+            clp::ir::epoch_time_ms_t timestamp,
+            clp::ffi::value_int_t utc_offset
```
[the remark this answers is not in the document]
junhaoliao: This is just a suggestion that the return type could be changed to return our internal presentation (e.g., a chrono object) of the offset. Note the underlying type of `clp::UtcOffset` is std::chrono::milliseconds which does not match our intended output unit.

src/clp_ffi_js/ir/StreamReader.hpp
```diff
@@ -263,21 +263,24 @@ auto StreamReader::generic_decode_range(
...
+                        "utcOffset": $5
                     });
                 },
                 results.as_handle(),
                 log_event_idx + 1,
                 log_level,
                 log_event_to_string(log_event).c_str(),
-                timestamp
+                timestamp,
+                utc_offset
```
junhaoliao, line 283: depending on the return type of `log_event_with_filter_data.get_utc_offset()`, below casting might not be necessary. the suggestion here is to say we can store the offset how ever we want, but the representation can be converted into the desired type & unit when needed. 
```suggestion
                std::chrono::duration_cast<std::chrono::minutes>(utc_offset).count()
```

junhaoliao (changes requested):
The structured is good. Let's make a few changes:
1. In the LogEventWithFilterData, let's store the offset as a chrono object so we will get more flexibility in manipulating the value in our implementation.
2. It might be better to always return the value in minutes, since that's a common unit for UTC offsets.
3. I made some suggestions about changing some internal representation of the offset to use `clp::UtcOffset`, but the underlying type is `std::chrono::milliseconds` which does not match `2.` above. Although it's also fine to store the value in milliseconds and eventually convert to minutes, i think we can create our own type `clp_ffi_js::ir::UtcOffset = std::chrono::minutes` and use it for the offsets' internal representations.

src/clp_ffi_js/ir/StreamReader.cpp
```diff
@@ -71,7 +72,8 @@ EMSCRIPTEN_BINDINGS(ClpStreamReader) {
             .value("STRUCTURED", clp_ffi_js::ir::StreamType::Structured)
             .value("UNSTRUCTURED", clp_ffi_js::ir::StreamType::Unstructured);
     emscripten::register_type<clp_ffi_js::ir::DecodedResultsTsType>(
-            "Array<{logEventNum: number, logLevel: number, message: string, timestamp: bigint}> "
+            "Array<{logEventNum: number, logLevel: number, message: string, timestamp: bigint, "
+            "utcOffsetMinutes: bigint}> "
```
junhaoliao, line 76: need to make this consistent with the rest of the code. personally i think it's fine to just use "utcOffset" since `minute` is a common unit for that.

src/clp_ffi_js/ir/StructuredIrUnitHandler.cpp
```diff
@@ -274,4 +291,44 @@ auto StructuredIrUnitHandler::get_timestamp(StructuredLogEvent const& log_event)
...
+
+    auto const& utc_offset_value{optional_utc_offset_value.value()};
+    if (false == utc_offset_value.is<clp::ffi::value_int_t>()) {
+        SPDLOG_ERROR(
+                "Protocol error: The utc offset must be a valid integer. Log event index: {}",
+                m_deserialized_log_events->size()
+        );
+        return cDefaultUtcOffset;
+    }
+    return UtcOffset{utc_offset_value.get_immutable_view<clp::ffi::value_int_t>()};
```
junhaoliao, line 332:
```suggestion
    // TODO: Support parsing of other time units, e.g., minutes, hours, etc..
    auto const utc_offset_seconds
            = std::chrono::seconds{utc_offset_value.get_immutable_view<clp::ffi::value_int_t>()};
    return std::chrono::duration_cast<UtcOffset>(utc_offset_seconds);
```

need to tell chrono that the input unit is seconds

junhaoliao (changes requested):
for the PR title, how about
```
feat: Add support for parsing UTC offset in seconds per log event and returning it in decode results (resolves #79).
```

src/clp_ffi_js/ir/StructuredIrUnitHandler.cpp
```diff
@@ -133,8 +133,9 @@ auto StructuredIrUnitHandler::handle_log_event(StructuredLogEvent&& log_event)
         -> clp::ffi::ir_stream::IRErrorCode {
     auto const timestamp = get_timestamp(log_event);
```
junhaoliao, line 134: In other places, the list initialization `{}` approach helps prevent narrowing conversions. In all places we need to consistently use `{}`, rather than `=`, for assignments whenever possible.

this is fine to leave in the current PR

@coderabbitai create an issue to track the refactoring request

junhaoliao:
hey @hoophalab could you upload the file used for validation and update the Validation performed section in the PR description?

PR #115, opened 2026-02-02
test: Add Vitest and ESLint infrastructure.

junhaoliao:
# Description

Adds the foundational build infrastructure for JavaScript/TypeScript testing and linting:

- Add Vitest as the test framework with Node.js configuration.
- Add ESLint with TypeScript support using yscope-dev-utils lint configs.
- Add TypeScript configuration (`tsconfig.json`) for type checking test files.
- Add npm scripts for linting (`lint:check:js`, `lint:fix:js`).
- Add task definitions for JS linting (`lint:check-js`, `lint:fix-js`) and testing (`test:js`).
- Update `.gitignore` to exclude `node_modules/` and `/test/data/` (cached test data).

This PR is part 1 of 6 in the test infrastructure series prototyped at #114.

# Checklist

  breaking change.

# Validation performed

1. Verified `task test` auto-installs dependencies via the `deps:node-modules` task:

   ```
   $ rm -rf node_modules
   $ task test
   task: [deps:node-modules] npm clean-install

   added 342 packages, and audited 343 packages in 933ms

   144 packages are looking for funding
     run `npm fund` for details

   found 0 vulnerabilities
   task: [test:js] npm run test

   > clp-ffi-js@0.6.1 test
   > vitest run

    RUN  v4.0.18 /home/junhao/workspace/clp-ffi-js

   No test files found, exiting with code 1
   ```

   The "No test files found" error is expected since test files are added in later PRs.

2. Verified `task lint:check-js` passes:

   ```
   $ rm -rf node_modules
   $ task lint:check-js
   task: Task "deps:node-modules" is up to date
   task: [lint:check-js] npm run lint:check:js

   > clp-ffi-js@0.6.1 lint:check:js
   > tsc --noEmit && eslint . --max-warnings 0

   [@stylistic/eslint-plugin]: You are using deprecated rule("jsx-sort-props"), please use
   "sort-jsx-props" in "eslint-plugin-perfectionist" instead.
   ```

   The deprecation warning is from an upstream eslint plugin and doesn't affect linting behavior.

## Summary by CodeRabbit

* **Chores**
  * Configured ESLint and TypeScript with stricter compiler and linting settings.
  * Added dev tooling and npm scripts for linting and testing, plus tasks for dependency setup and lint workflows.
  * Updated ignore rules to exclude dependencies and test data.

* **Tests**
  * Added Vitest with automated discovery and a 30s timeout.
  * Introduced a test task workflow to run the test suite.


taskfiles/test.yaml
```diff
@@ -0,0 +1,9 @@
+version: "3"
+
+tasks:
+  default:
+    deps: ["js"]
+
+  js:
+    deps: [":clp-ffi-js", ":deps:node-modules"]
+    cmd: "npm run test"
```
Bill-hbrhbr:
```suggestion
    cmd: "npm exec -- vitest run"
```
I think I prefer this way of invoking vitest due to its parity with `uv run`.

Bill-hbrhbr (changes requested):

vitest.config.ts
```diff
@@ -0,0 +1,10 @@
+import {defineConfig} from "vitest/config";
+
+
+export default defineConfig({
+    test: {
+        include: ["test/**/*.test.ts"],
```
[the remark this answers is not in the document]
Bill-hbrhbr: Also do you want to add exclusion for `test/data` like what you have for `.gitignore`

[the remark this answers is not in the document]
junhaoliao: do note the strategy for detecting files to be linted and test files to be run are different.

in the eslint config, we use a ignore (deny) list; here in the test config we use an include (allow) list. mixing the usage of allow and deny lists seems a bit unnecessary and confusing here, so i prefer not to do that

regarding excluding `test/data` - do note the `include` lists we use explicit file extension `.test.ts`. we don't expect to place any test in `test/data` unless someday we write self-generating code, so there shouldn't be any correctness issue with the current `include`. then we should not need to add an `ignore` for `test/data`, which again avoids the mixed usages of `include` and `ignores`

tsconfig.json
```diff
@@ -0,0 +1,34 @@
+{
+    "compilerOptions": {
```
junhaoliao: all [20 lint checks](https://www.typescriptlang.org/tsconfig/#:~:text=%22compilerOptions%22-,Type%20Checking,-allowUnreachableCode%2C) are now active. Here's the breakdown:

Explicitly set:
1. strict
2. exactOptionalPropertyTypes
3. noFallthroughCasesInSwitch
4. noImplicitOverride
5. noImplicitReturns
6. noPropertyAccessFromIndexSignature
7. noUncheckedIndexedAccess
8. noUnusedLocals
9. noUnusedParameters

Set to false which means "error":
1. allowUnreachableCode
2. allowUnusedLabels 

These two have inverted semantics compared to the others. Their default is undefined (warn only), and setting them to false promotes warnings to errors.


Implied by strict: true
1. alwaysStrict
2. noImplicitAny
3. noImplicitThis
4. strictBindCallApply
5. strictBuiltinIteratorReturn
6. strictFunctionTypes
7. strictNullChecks
8. strictPropertyInitialization
9. useUnknownInCatchVariables

Bill-hbrhbr (approved):

tsconfig.json
junhaoliao: @Bill-hbrhbr suggested offline that this file may be moved to https://github.com/y-scope/yscope-dev-utils . we agree that it will be useful to maintain some tsconfig.json file as some reference there. even if we cannot import the config from yscope-dev-utils , at least the config file there can serve as some reference doc. I will do it the next time we start a new TypeScript project.

PR #144, opened 2026-04-07
fix(sfa)!: Remove top-level await from WASM initialization; Explicitly locate worker WASM path in the browser entry point.

Bill-hbrhbr:
# Description

This PR fixes two downstream integration issues between Tritonparse and clp-ffi-js.

1. The worker entrypoint currently relies on top level `await`, which is not a good lifecycle pattern and also breaks downstream production builds.
In Tritonparse, Vite fails when bundling the worker because top level `await` is not supported for the `iife `output format.
```
bingranhu@baker21:~/tritonparse/website$ npm run build

> tritonparse-website@0.4.0 build
> tsc -b && vite build

vite v8.0.3 building client environment for production...
✓ 1337 modules transformed.
✗ Build failed in 312ms
error during build:
Build failed with 1 error:

[UNSUPPORTED_FEATURE] Error: Top-level await is currently not supported with the 'iife' output format
    ╭─[ node_modules/clp-ffi-js/dist/clp_ffi_js/sfa/index-worker.js:12:11 ]
    │
 12 │ setModule(await mainModuleFactory());
    │           ────────────┬────────────
    │                       ╰──────────────
────╯
l
    at aggregateBindingErrorsIntoJsError (file:///home/bingranhu/tritonparse/website/node_modules/rolldown/dist/shared/error-BLhcSyeg.mjs:48:18)
    at unwrapBindingResult (file:///home/bingranhu/tritonparse/website/node_modules/rolldown/dist/shared/error-BLhcSyeg.mjs:18:128)
    at #build (file:///home/bingranhu/tritonparse/website/node_modules/rolldown/dist/shared/rolldown-build-CPrIX9V6.mjs:3313:34)
    at async buildEnvironment (file:///home/bingranhu/tritonparse/website/node_modules/vite/dist/node/chunks/node.js:32849:64)
    at async Object.build (file:///home/bingranhu/tritonparse/website/node_modules/vite/dist/node/chunks/node.js:33271:19)
    at async Object.buildApp (file:///home/bingranhu/tritonparse/website/node_modules/vite/dist/node/chunks/node.js:33268:153)
    at async CAC.<anonymous> (file:///home/bingranhu/tritonparse/website/node_modules/vite/dist/node/cli.js:778:3) {
  errors: [Getter/Setter]
}
```
2. Failures to resolve the worker WASM module currently manifest “magic number” errors. In the failing case, the logs show that the fetched response had the wrong MIME type, then contained HTML bytes like <!DO instead of WASM bytes \0asm, which indicates that the worker WASM URL was wrong or unresolved.
```
dataLoader.ts:494 wasm streaming compile failed: TypeError: Failed to execute 'compile' on 'WebAssembly': Incorrect response MIME type. Expected 'application/wasm'.
dataLoader.ts:494 falling back to ArrayBuffer instantiation
dataLoader.ts:494 failed to asynchronously prepare wasm: CompileError: WebAssembly.instantiate(): expected magic word 00 61 73 6d, found 3c 21 44 4f @+0
dataLoader.ts:494 Aborted(CompileError: WebAssembly.instantiate(): expected magic word 00 61 73 6d, found 3c 21 44 4f @+0)
dataLoader.ts:511 Error decompressing or parsing clp stream: RuntimeError: Aborted(CompileError: WebAssembly.instantiate(): expected magic word 00 61 73 6d, found 3c 21 44 4f @+0). Build with -sASSERTIONS for more info.
```

# Checklist

  breaking change.

# Validation performed

* [X] Copied from `beta/sfa` and verified to resolve downstream issues.

## Release Notes

* **Refactor**
  * Archive reader initialization is now asynchronous, allowing for deferred WASM module loading and improved error handling.
  * Module initialization changed from eager to lazy loading, enabling retry logic on failure and better resource management across different environments.
  * Tests added to verify single-flight concurrency handling and factory initialization constraints.


src/clp_ffi_js/sfa/index-worker.ts
```diff
@@ -1,17 +1,24 @@
...
 
 import mainModuleFactory from "#clp-ffi-js/worker";
+import workerWasmUrl from "#clp-ffi-js/worker-wasm?url";
+
 
+setModuleFactory(() => {
+    return mainModuleFactory({
+        locateFile: (path: string) => {
+            if ("ClpFfiJs-worker.wasm" === path) {
+                return workerWasmUrl;
```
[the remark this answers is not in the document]
Bill-hbrhbr: `locateFile` will be improved when we decide on where to actually host the WASM module.

src/clp_ffi_js/sfa/index-worker.ts
```diff
@@ -1,17 +1,24 @@
...
+            }
 
-setModule(await mainModuleFactory());
+            return path;
+        },
+    });
+});
```
hoophalab: I guess this doesn't work on their system? Do you know why?
```suggestion


setModuleFactory(mainModuleFactory);
```

hoophalab (commented):
LGTM in general. One question

src/clp_ffi_js/sfa/index-worker.ts
Bill-hbrhbr: If `locateFile` returns only the basename `ClpFfiJs-worker.wasm`, then the browser resolves that as a relative URL against the current browser page, not against the package file that requested it.

In an app like this, that usually means the request goes to something like:
```
http://<host>:<port>/ClpFfiJs-worker.wasm
```
But that file is not actually served there. The server then often falls back to returning the app HTML page, typically `index.html`, for unknown routes. So the fetch succeeds, but the response body is HTML, not a WASM binary.

That is why WebAssembly instantiation fails with:
```
  - expected WASM magic bytes: 00 61 73 6d
  - got HTML bytes: 3c 21 44 4f (<!DO)
```

What we are doing is basically helping guide the server to the actual location of the WASM binary.

src/clp_ffi_js/sfa/index-worker.ts
```diff
@@ -1,17 +1,24 @@
 /**
  * SFA module entry point for worker and browser environments.
  *
```
junhaoliao, line 3:
```suggestion
 *
 * NOTE: This entry point uses the `?url` import suffix to resolve the `.wasm` asset, which is a
 * Vite/Rollup-specific convention and is not recognized by every bundler (e.g., Webpack, esbuild).
 * Consumers using other bundlers may need to provide their own entry point.
```

src/clp_ffi_js/sfa/index-worker.ts
```diff
@@ -1,17 +1,24 @@
...
  * @module
  */
-import {setModule} from "./module.js";
+import {setModuleFactory} from "./module.js";
 
 import mainModuleFactory from "#clp-ffi-js/worker";
+import workerWasmUrl from "#clp-ffi-js/worker-wasm?url";
+
 
+setModuleFactory(() => {
```
junhaoliao, line 45:
```suggestion
// Emscripten's default `locateFile` resolves the `.wasm` file against the current page URL, which
// fails when a bundler emits the `.wasm` under a hashed asset path separate from the JS bundle.
// Override it to return the bundler-resolved URL instead.
setModuleFactory(() => {
```

src/clp_ffi_js/sfa/module.ts
```diff
@@ -1,36 +1,41 @@
...
- * @throws {Error} If the module has already been set.
+ * @param newMainModuleFactory Factory for loading the WASM module.
+ * @throws {Error} If the factory has already been set.
  */
-const setModule = (newMainModule: MainModule): void => {
-    if (null !== mainModule) {
-        throw new Error("WASM module has already been initialized.");
+const setModuleFactory = (newMainModuleFactory: () => Promise<MainModule>): void => {
+    if (null !== mainModuleFactory) {
+        throw new Error("WASM module factory has already been initialized.");
```
junhaoliao:
```suggestion
        throw new Error("WASM module factory has already been set.");
```

src/clp_ffi_js/sfa/module.ts
```diff
@@ -1,36 +1,41 @@
...
-    if (null === mainModule) {
-        throw new Error("WASM module has not been initialized.");
+const getModule = async (): Promise<MainModule> => {
+    if (null === mainModulePromise) {
+        if (null === mainModuleFactory) {
+            throw new Error("WASM module factory has not been initialized.");
+        }
+
+        mainModulePromise = mainModuleFactory();
     }
```
junhaoliao: 1. use early return
2. allow retry on rejected Promises instead of caching the dead rejected promise
```suggestion
    if (null !== mainModulePromise) {
        return mainModulePromise;
    }
    if (null === mainModuleFactory) {
        throw new Error("WASM module factory has not been set.");
    }

    const promise = mainModuleFactory();

    // Clear the cached promise on rejection so the next caller retries.
    promise.catch(() => {
        if (mainModulePromise === promise) {
            mainModulePromise = null;
        }
    });
    mainModulePromise = promise;
```

junhaoliao (changes requested):
i think we should also add some unit tests to cover the module import usages.

The tests must verify the full lifecycle contract of setModuleFactory / getModule - including error cases and cache/retry semantics - and must run in all four Vitest projects (node, browser (chromium), browser (firefox), browser (webkit)).

here are the contract to verify
1. Factory-not-set error: calling getModule() before setModuleFactory(...) must throw an error whose message contains the substring factory has not been set (or similar - match on a regex).
2. Single-flight concurrency: if getModule() is called N times concurrently (before the factory's returned promise resolves), the factory must be invoked exactly once, and all N callers must receive the same resolved module instance.
3. Promise caching: after a successful resolution, subsequent getModule() calls must continue to return the same resolved module (the cached promise is reused).
4. Retry-on-rejection: if the factory's first invocation rejects, the cached promise must be cleared so that the next getModule() call re-invokes the factory. The test should use a mock factory that rejects once, then resolves and assert:
    - the first getModule() call rejects with the thrown error
    - the second call resolves with the stub module
    - the factory was called exactly twice
5. Double-set error: calling setModuleFactory(...) a second time (after it has already been called once) must throw an error whose message contains already been set (or similar).


here's what i came up with as i review the code
```ts
import type {MainModule} from "clp-ffi-js/node";
import {
    describe,
    expect,
    it,
    vi,
} from "vitest";


type ModuleApi = typeof import("../src/clp_ffi_js/sfa/module.js");

let freshModuleCounter = 0;

/**
 * Imports `module.ts` with fresh singleton state. Uses a cache-busting query string so each call
 * yields a distinct module instance; this works across Node and browser, whereas
 * `vi.resetModules()` alone does not reliably reset module-level state in the browser.
 *
 * @return A fresh copy of the module's exports.
 */
const loadFreshModule = async (): Promise<ModuleApi> => {
    freshModuleCounter += 1;
    const specifier = `../src/clp_ffi_js/sfa/module.js?v=${freshModuleCounter.toString()}`;

    // eslint-disable-next-line no-inline-comments
    return (await import(/* @vite-ignore */ specifier)) as ModuleApi;
};

describe("sfa/module.ts", () => {
    describe("getModule", () => {
        it("throws if the factory has not been set", async () => {
            const {getModule} = await loadFreshModule();

            expect(() => getModule()).toThrow(/factory has not been set/);
        });

        it("invokes the factory exactly once across concurrent callers", async () => {
            const {setModuleFactory, getModule} = await loadFreshModule();

            const stubModule = {} as MainModule;
            const factory = vi.fn(() => Promise.resolve(stubModule));
            setModuleFactory(factory);

            const [r1,
                r2,
                r3] = await Promise.all([
                getModule(),
                getModule(),
                getModule(),
            ]);

            expect(factory).toHaveBeenCalledTimes(1);
            expect(r1).toBe(stubModule);
            expect(r2).toBe(stubModule);
            expect(r3).toBe(stubModule);
        });

        it("returns the same cached promise on subsequent calls", async () => {
            const {setModuleFactory, getModule} = await loadFreshModule();

            const stubModule = {} as MainModule;
            setModuleFactory(() => Promise.resolve(stubModule));

            await expect(getModule()).resolves.toBe(stubModule);
            await expect(getModule()).resolves.toBe(stubModule);
        });

        it("clears the cached promise on rejection so the next caller retries", async () => {
            const {setModuleFactory, getModule} = await loadFreshModule();

            const stubModule = {} as MainModule;
            const factory = vi.fn<() => Promise<MainModule>>()
                .mockRejectedValueOnce(new Error("transient failure"))
                .mockResolvedValueOnce(stubModule);

            setModuleFactory(factory);

            await expect(getModule()).rejects.toThrow("transient failure");
            await Promise.resolve();
            await expect(getModule()).resolves.toBe(stubModule);
            expect(factory).toHaveBeenCalledTimes(2);
        });
    });

    describe("setModuleFactory", () => {
        it("throws if called more than once", async () => {
            const {setModuleFactory} = await loadFreshModule();

            const noopFactory = (): Promise<MainModule> => Promise.resolve({} as MainModule);
            setModuleFactory(noopFactory);

            expect(() => {
                setModuleFactory(noopFactory);
            }).toThrow(/already been set/);
        });
    });
});
```

src/clp_ffi_js/sfa/module.ts
```diff
@@ -1,36 +1,50 @@
...
 };
 
 /**
  * Returns the shared WASM module.
  *
- * @return The loaded module.
- * @throws {Error} If the module has not been set via {@link setModule}.
+ * @return A promise that resolves to the loaded module.
+ * @throws {Error} If the module factory has not been set via {@link setModuleFactory}.
  */
```
junhaoliao, line 26:
```suggestion
 * @throws {Error} Propagates `mainModuleFactory`'s exceptions.
 */
```

test/sfaModule.test.ts
```diff
@@ -0,0 +1,125 @@
...
+                getModule(),
+                getModule(),
+            ];
+
+            expect(factory).toHaveBeenCalledTimes(1);
+
+            deferred.resolve(stubModule);
+            const [result1,
+                result2,
+                result3] = await Promise.all(pendingResults);
```
junhaoliao:
```suggestion
            const [
                result1,
                result2,
                result3
            ] = await Promise.all(pendingResults);
```

junhaoliao (changes requested):
just some really nitpicking suggestions

---

i think this is a breaking change and therefore the title should include `!`
```
fix(sfa)!: Remove top-level await from WASM initialization; Explicitly locate worker WASM path in the browser entry point.
```

junhaoliao (approved):

junhaoliao:
approved and fixed some lint error

- discussed offline that i'll perform the merge since my change is only lint-related

PR #147, opened 2026-04-17
feat(sfa): Add utilities for checking CLP JSON single-file archive magic bytes.

Bill-hbrhbr:
# Description

As title. This is useful for downstream projects that need to branch on archive format before deciding how to load or decode the input.

# Checklist

  breaking change.

# Validation performed

* [X] CI and new unit tests pass.

## Summary by CodeRabbit

* **New Features**
  * Added reliable single-file archive detection using a fixed header check and exposed utilities to make archive validation easier across the module.

* **Tests**
  * Added automated tests validating detection behavior and boundary conditions (partial views, offsets, empty and non-matching inputs) to ensure correctness.


src/clp_ffi_js/sfa/index.ts
```diff
@@ -1,2 +1,5 @@
 export {ClpArchiveReader} from "./ClpArchiveReader.js";
 export type {FileInfo} from "./types.js";
+export {
+    CLP_SFA_MAGIC_BYTES, isClpFile,
```
junhaoliao:
```suggestion
    CLP_SFA_MAGIC_BYTES,
    isClpFile,
```

src/clp_ffi_js/sfa/utils.ts
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Starting byte sequence that identifies a CLP JSON single-file archive (SFA).
+ */
+const CLP_SFA_MAGIC_BYTES = [
+    CLP_SFA_MAGIC_BYTE_0,
+    CLP_SFA_MAGIC_BYTE_1,
+    CLP_SFA_MAGIC_BYTE_2,
+    CLP_SFA_MAGIC_BYTE_3,
+] as const;
```
junhaoliao: appreciate the attempt in avoiding lint rules suppression but this may seems cleaner
```suggestion
/**
 * Starting byte sequence that identifies a CLP JSON single-file archive (SFA).
 */
const CLP_SFA_MAGIC_BYTES = [
    // eslint-disable-next-line no-magic-numbers, @stylistic/array-element-newline
    0xFD, 0x2F, 0xC5, 0x30,
] as const;
```

src/clp_ffi_js/sfa/utils.ts
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Detects whether a buffer is a CLP JSON single-file archive (SFA).
+ *
+ * @param buffer Buffer containing the archive bytes.
+ * @return `true` if the buffer starts with the CLP SFA magic bytes.
+ */
+const isClpFile = (buffer: ArrayBuffer): boolean => {
```
junhaoliao: forcing users to pass in an `ArrayBuffer` seems too restrictive. While it's possible to get a buffer from an `Uint8Array`, usages can easily go wrong depending on how the Uint8Array is created. e.g., in some user code
```ts
// create a view of someLargeBuffer's [100, length) bytes, assuming the first 100 bytes in this buffer is garbage data
const view = new Uint8Array(someLargeBuffer).subarray(100);
// `view.buffer` refers to the `someLargeBuffer` - comparing the bytes in the garbage data part with magic number gives unexpected results
console.log(isClpFile(view.buffer));
```

junhaoliao: we may consider something like ... see the below `const header =` line change as well

```suggestion
const isClpFile = (input: ArrayBuffer | ArrayBufferView): boolean => {
    let buffer: ArrayBuffer;
    let offset = 0;
    let length: number;

    if (input instanceof ArrayBuffer) {
        buffer = input;
        length = input.byteLength;
    } else {
        buffer = input.buffer;
        offset = input.byteOffset;
        length = input.byteLength;
    }
```

src/clp_ffi_js/sfa/utils.ts
```diff
@@ -0,0 +1,37 @@
...
+    if (buffer.byteLength < CLP_SFA_MAGIC_BYTES.length) {
+        return false;
+    }
+    const header = new Uint8Array(buffer, 0, CLP_SFA_MAGIC_BYTES.length);
```
junhaoliao:
```suggestion
    const header = new Uint8Array(buffer, offset, CLP_SFA_MAGIC_BYTES.length);
```

src/clp_ffi_js/sfa/utils.ts
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Detects whether a buffer is a CLP JSON single-file archive (SFA).
+ *
+ * @param buffer Buffer containing the archive bytes.
+ * @return `true` if the buffer starts with the CLP SFA magic bytes.
+ */
+const isClpFile = (buffer: ArrayBuffer): boolean => {
+    if (buffer.byteLength < CLP_SFA_MAGIC_BYTES.length) {
+        return false;
```
junhaoliao, line 21: let's also add coverage for this branch. e.g.,
```
it("should reject buffers that don't start with the CLP SFA magic bytes", () => {
    expect(isClpFile(new ArrayBuffer(0))).toBe(false);
    // ... 
});
```

junhaoliao (changes requested):
some very minor nits

for the title, how about:
```
feat(sfa): Add utilities for checking CLP JSON single-file archive magic bytes.
```

src/clp_ffi_js/sfa/utils.ts
```diff
@@ -0,0 +1,30 @@
...
+    0xFD, 0x2F, 0xC5, 0x30,
+] as const;
+
+/**
+ * Detects whether a buffer is a CLP JSON single-file archive (SFA).
+ *
+ * @param input Buffer or view containing the archive bytes.
+ * @return `true` if the input starts with the CLP SFA magic bytes.
+ */
+const isClpFile = (input: ArrayBuffer | ArrayBufferView): boolean => {
```
junhaoliao: sorry for missing this earlier - do you think it would be more appropriate to name this `isClpSfaFile`?

junhaoliao (changes requested):
should be good to go once this is addressed - https://github.com/y-scope/clp-ffi-js/pull/147/changes#r3152265746

src/clp_ffi_js/sfa/utils.ts
Bill-hbrhbr: Replaced with an even more explicit name

Bill-hbrhbr: `isClpJsonSingleFileArchive`

junhaoliao (approved):
lgtm

PR #155, opened 2026-07-14
feat(sfa): Add API for decoding all log events.

Bill-hbrhbr:
# Description

Adds the native decodeAll binding and returns raw event objects across the WASM boundary. ClpArchiveReader wraps each raw event in a public LogEvent class with read-only accessors and getKvPairs() support for parsing structured JSON messages.

# Checklist

  breaking change.

# Validation performed

* [X] lint and unit tests. pass.

## Summary by CodeRabbit

- **New Features**
  - Added decoding of all archive log events in a single call.
  - Introduced a `LogEvent` object model (index, timestamp, message) and exposed it publicly.
  - Added optional JSON parsing for log event messages, including runtime validation that the parsed value is a JSON object.
  - Exported shared JSON typing utilities (`JsonValue`, `JsonObject`, and `RawLogEvent`) and a new `isJsonObject` helper for downstream use.

### DIFF
--- a/src/clp_ffi_js/sfa/ClpArchiveReader.ts
+++ b/src/clp_ffi_js/sfa/ClpArchiveReader.ts
@@ -75,13 +75,7 @@ class ClpArchiveReader {
     decodeAll (): LogEvent[] {
         return this.#getWasmReader()
             .decodeAll()
-            .map((rawEvent) => {
-                return new LogEvent(
-                    rawEvent.logEventIdx,
-                    rawEvent.timestamp,
-                    rawEvent.message
-                );
-            });
+            .map((rawEvent) => new LogEvent(rawEvent));
     }
 
     /**
--- a/src/clp_ffi_js/sfa/ClpArchiveReader.ts
+++ b/src/clp_ffi_js/sfa/ClpArchiveReader.ts
@@ -1,3 +1,4 @@
+import {LogEvent} from "./LogEvent.js";
 import {getModule} from "./module.js";
 import type {FileInfo} from "./types.js";
 
@@ -63,6 +64,24 @@ class ClpArchiveReader {
      */
     getFileInfos (): FileInfo[] {
         return this.#getWasmReader().getFileInfos();
+    }
+
+    /**
+     * Decodes all log events in global log-event-index order.
+     *
+     * @return Decoded log events.
+     * @throws {Error} If the reader has been closed.
+     */
+    decodeAll (): LogEvent[] {
+        return this.#getWasmReader()
+            .decodeAll()
+            .map((rawEvent) => {
+                return new LogEvent(
+                    rawEvent.logEventIdx,
+                    rawEvent.timestamp,
+                    rawEvent.message
+                );
+            });
     }
 
     /**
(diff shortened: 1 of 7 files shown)


src/clp_ffi_js/sfa/ClpArchiveReader.ts
```diff
@@ -65,6 +66,18 @@ class ClpArchiveReader {
         return this.#getWasmReader().getFileInfos();
     }
 
+    /**
+     * Decodes all log events in global log-event-index order.
+     *
+     * @return Decoded log events.
+     * @throws {Error} If the reader has been closed.
+     */
+    decodeAll (): LogEvent[] {
```
junhaoliao, line 75: as discussed offline - can we add coverage for this new public WASM path? something like

```ts
const events = reader.decodeAll();
expect(events).toHaveLength(Number(CLP_JSON_TEST_LOG_FILES_EXPECTED_EVENT_COUNT));
events.forEach((event, index) => {
    expect(event).toBeInstanceOf(LogEvent);
    expect(event.logEventIdx).toBe(BigInt(index));
    expect(typeof event.timestamp).toBe("bigint");
    expect(typeof event.message).toBe("string");
});
expect(reader.decodeAll()).toEqual(events);
```

src/clp_ffi_js/sfa/ClpArchiveReader.ts
```diff
@@ -65,6 +66,18 @@ class ClpArchiveReader {
         return this.#getWasmReader().getFileInfos();
     }
 
+    /**
+     * Decodes all log events in global log-event-index order.
+     *
+     * @return Decoded log events.
+     * @throws {Error} If the reader has been closed.
```
junhaoliao, line 73:
```suggestion
     * @throws {Error} If the reader has been closed or the archive cannot be decoded.
```

src/clp_ffi_js/sfa/LogEvent.ts
```diff
@@ -0,0 +1,55 @@
...
+ */
+class LogEvent implements RawLogEvent {
+    /**
+     * Global log event index.
+     */
+    declare readonly logEventIdx: bigint;
+
+    /**
+     * Epoch timestamp.
+     */
```
junhaoliao, line 20: would this be more clear?

```suggestion
    /**
     * Timestamp in milliseconds since the Unix epoch.
     */
```

src/clp_ffi_js/sfa/SfaReader.cpp
```diff
@@ -66,13 +66,40 @@ auto SfaReader::get_file_infos() const -> FileInfoArrayTsType {
...
+    auto decoded_result{m_reader.decode_all()};
+    if (decoded_result.has_error()) {
+        auto const error{decoded_result.error()};
+        auto const err_msg{fmt::format(
+                "Failed to decode SFA archive: {} - {}.",
+                error.category().name(),
+                error.message()
+        )};
+        SPDLOG_ERROR("{}", err_msg);
+        throw std::runtime_error{err_msg};
```
junhaoliao, line 80: shall we use `ClpFfiJsException` for this new FFI failure path?
```suggestion
        throw clp_ffi_js::ClpFfiJsException{
                clp::ErrorCode::ErrorCode_Failure,
                __FILENAME__,
                __LINE__,
                err_msg
        };
```

junhaoliao (changes requested):
lgtm

as discussed offline, it would be better if we can add some unit test to cover the changes

